PR #32, opened 2023-06-14
HNSW

usamoi:


.cargo/config.toml
```diff
@@ -1,3 +1,7 @@
+[build]
+rustflags = ["-Ctarget-cpu=native"]
```
VoVAllen, line 2: Will pgvecto.ra be distributed as binary or only source code, if user install this as extension?

README.md
```diff
@@ -71,42 +79,82 @@ CREATE EXTENSION vectors;
...
-- `<=>`: cosine distance
+- `<#>`: negative dot product distance
+- `<=>`: negative square cosine distance
 
 ```sql
 -- call the distance function through operators
 
 -- square Euclidean distance
-SELECT array[1, 2, 3] <-> array[3, 2, 1];
+SELECT '[1, 2, 3]' <-> '[3, 2, 1]';
```
VoVAllen, line 89: Does '[1,2,3]' mean a string here? Will there be extra parsing cost?

README.md
```diff
@@ -71,42 +79,82 @@ CREATE EXTENSION vectors;
...
+nlist = 1000
+nprobe = 10
+$$);
+```
+
+The index must be built on a vector column. Failure to match the actual vector dimension with the dimension type modifier may result in an unsuccessful index building.
+
+The operator class determines the type of distance measurement to be used. At present, `l2_ops`, `dot_ops`, and `cosine_ops` are supported.
+
+You can specify the indexing and the vectors to be stored in the disk by setting `storage_vectors = "disk"`, and `storage = "disk"`. On this condition, `size_disk` must be specified.
```
VoVAllen, line 149: What does storage on disk vs ram mean? Is this part of the underlying hnsw setting or a general setting for all indexes?

src/utils/vec2.rs
```diff
@@ -0,0 +1,67 @@
+use crate::prelude::*;
+use std::ops::{Deref, DerefMut, Index, IndexMut};
+
+#[derive(Debug, Clone)]
+pub struct Vec2 {
+    dims: u16,
+    v: Box<[Scalar]>,
+}
+
+impl Vec2 {
```
VoVAllen, line 10: What does vec2 mean?

src/utils/semaphore.rs
```diff
@@ -0,0 +1,47 @@
+use std::ops::{Deref, DerefMut};
+
+pub struct Semaphore<T> {
```
VoVAllen, line 3: Where did you use the semaphore?

README.md
usamoi: `'[1, 2, 3]'` will be casted to `vector`, since operands of `<->` are `vector`.

Yes, it will. A cast function from an array to `vector` might be helpful.

src/utils/semaphore.rs
usamoi: It's used in HNSW.

src/utils/vec2.rs
usamoi: Array of vectors.

README.md
usamoi: It means whether the indexing is stored in memory or in an memory-mapped file. `storage` is an option for both HNSW and IVF.

.cargo/config.toml
usamoi: Source code, I think.

README.md
```diff
@@ -71,42 +79,82 @@ CREATE EXTENSION vectors;
...
+CREATE TABLE items (id bigserial PRIMARY KEY, emb vector(3));
 -- insert values
-INSERT INTO items (emb) VALUES (ARRAY[1,2,3]), (ARRAY[4,5,6]);
+INSERT INTO items (emb) VALUES ('[1,2,3]'), ('[4,5,6]');
 -- query the similar embeddings
-SELECT * FROM items ORDER BY emb <-> ARRAY[3,2,1]::real[] LIMIT 5;
+SELECT * FROM items ORDER BY emb <-> '[3,2,1]' LIMIT 5;
 -- query the neighbors within a certain distance
-SELECT * FROM items WHERE emb <-> ARRAY[3,2,1]::real[] < 5;
+SELECT * FROM items WHERE emb <-> '[3,2,1]' < 5;
```
VoVAllen, line 110: Can we still support `ARRAY[3,2,1]::real[]` like grammar? And what's the error message now if user do `SELECT * FROM items WHERE emb <-> '[a,2,1]' < 5; That the string cannot be parsed
`

README.md
VoVAllen: What's the purpose of memory-mapped file? Can we recover the index from this file? When does the recover may happen?

README.md
usamoi: Now the message is `Bad charactor with ascii 0x61.`. Since ASCII of `a` is `0x61`.

usamoi: (For the question below.) Yes, we can recover the index from the memory-mapped file. If a structure needs to allocate space in the memory-mapped file, the root pointer of the structure is stored so that we can recover the structure from the pointer.

VoVAllen:
@all-contributors please add @usamoi for code contribution.

PR #38, opened 2023-07-31
feat: add ci for nightly release

usamoi:
Nightly releases will be built on 00:00 UTC+8 if there are any commits in the past 24 hours.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,89 @@
+name: Release
+
+on:
+  schedule:
+    # 00:00 UTC+8 -> 16:00
+    - cron: "0 16 * * *"
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Sets
+        id: sets
+        env:
+          TZ: Asia/Shanghai
+        run: |
+          commit_date=$(git log -1 --since="24 hours ago" --pretty=format:"%cI")
+          if [[ -n "$commit_date" ]];
+          then echo "recent=true" >> $GITHUB_OUTPUT;
+          else echo "recent=false" >> $GITHUB_OUTPUT;
+          fi
+          echo "now=$(date +'%Y%m%d')" >> $GITHUB_OUTPUT;
+        shell: bash
+    outputs:
+      recent: ${{ steps.sets.outputs.recent }}
+      now: ${{ steps.sets.outputs.now }}
+  release:
+    permissions:
+      contents: write
+      pull-requests: read
+    strategy:
+      matrix:
+        version: [15]
+    needs: setup
+    runs-on: ubuntu-latest
+    if: ${{ needs.setup.outputs.recent == 'true' }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


gaocegege (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+          touch ./target/vectors_0.0-0_amd64/DEBIAN/postinst
+          chmod +x ./target/vectors_0.0-0_amd64/DEBIAN/postinst
+          echo "Package: vectors" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "Version: 0.0" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "Architecture: amd64" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "Maintainer: example@example.com" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "Description: No descrption." >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "su postgres -c \"psql -U postgres -c 'ALTER SYSTEM SET shared_preload_libraries = \"vectors\"'\"" >> ./target/vectors_0.0-0_amd64/DEBIAN/postinst
+          mv ./target/release/vectors-pg$(psql --version | awk -F"[ ,.]" '{print $3}')/usr ./target/vectors_0.0-0_amd64/usr
+          dpkg-deb --build --root-owner-group ./target/vectors_0.0-0_amd64
```
VoVAllen, line 69: We should support dynamic version

.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          cargo pgrx package
+          # deb
+          mkdir ./target/vectors_0.0-0_amd64
+          mkdir ./target/vectors_0.0-0_amd64/DEBIAN
+          touch ./target/vectors_0.0-0_amd64/DEBIAN/control
+          touch ./target/vectors_0.0-0_amd64/DEBIAN/postinst
+          chmod +x ./target/vectors_0.0-0_amd64/DEBIAN/postinst
+          echo "Package: vectors" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
+          echo "Version: 0.0" >> ./target/vectors_0.0-0_amd64/DEBIAN/control
```
VoVAllen, line 63: ditto

VoVAllen (changes requested):
Please change the trigger to the new tags, and use tag's version as the extension's version.

Example (https://github.com/tensorchord/envd/blob/main/.github/workflows/release.yml):
```yaml
on:
  push:
    tags:
      - 'v*'
``` 
and use `${{ github.event.release.tag_name }}` to get the version


Also better to change the building process to docker instead of pure github action. This makes local debug and build much easier

usamoi:
@VoVAllen I'm not sure when will the version move from 0.0.0...

VoVAllen:
@usamoi I already made one https://github.com/tensorchord/pgvecto.rs/releases/tag/v0.1.0. However I think it should be 0.0.1 if we're not confident now. We can start with 0.0.1, it's fine. What do you think?

usamoi:
r?

gaocegege (commented):
@VoVAllen

VoVAllen (approved):
Let's merge this first to see whether it works

PR #42, opened 2023-08-04
fix: outdated readme and envd

usamoi:


README.md
```diff
@@ -16,84 +16,130 @@ pgvecto.rs is a Postgres extension that provides vector similarity search functi
...
+```
+
+### Install Postgresql and pgrx
+
 ```sh
+sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
+wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
+sudo apt-get update
+sudo apt-get -y install libpq-dev postgresql-15 postgresql-server-dev-15
 cargo install cargo-pgrx --git https://github.com/tensorchord/pgrx.git --rev $(cat Cargo.toml | grep "pgrx =" | awk -F'rev = "' '{print $2}' | cut -d'"' -f1)
```
VoVAllen, line 45: Can we update to official pgrx beta version since the header file is included?

usamoi: @VoVAllen I'm not sure if we need include more in the future.

VoVAllen:
Thanks!

PR #45, opened 2023-08-11
Update sql in readme

yeya24:
Following the example in readme, there is no column called `emb` so updated it to the correct name.


kemingy (approved):
Thanks for your contribution! 🎉

gaocegege:
@all-contributors Please add @yeya24 for the doc!

PR #46, opened 2023-08-14
prefilter

usamoi:


gaocegege:
Please rebase the upstream master, there are many conflicts.

kemingy:
May I know why we need the prefilter? AFAIK, the prefilter cannot reuse the HNSW indexing. I think we need the post filter.

PR #47, opened 2023-08-14
feat: prefilter

usamoi:
Pre-filter is enabled by default and no more attention should be paid. It guarantees that indexes will always return enough KNN search results, even if there is a WHERE clause and transaction isolation prevents access of some items.


usamoi:
r?

gaocegege:
Build failed.

VoVAllen:
Thanks for your contribution! Please add some descriptions to this PR about what is done in this PR (quantization? filter?). And also how to use the filter feature?

PR #49, opened 2023-08-16
feat: use simd if cpu supports

AveryQi115:
# Major Changes
- add vec version of distance_squared_l2, length, distance_cosine, distance_dot
- add vectorization support check, if not supports, fall back to scalar compute

## Vectorization support check
All vectorized operations rely on 128 bit register (f32x4), target feature should support it for vectorized operation
- target_feature "neno" for arm and aarch64
- target_feature "sse","sse2","sse3","ssse3","sse4.1", "sse4.2","sse4a" for x86 and x86_64

# Test
- Unit Test in mod distance_tests, difference between vec and scalar operation is less than 1e-5

# Dependencies
- portable_simd https://github.com/rust-lang/portable-simd

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,9 @@
 #![feature(never_type)]
 #![allow(clippy::complexity)]
 #![allow(clippy::style)]
+#![feature(array_chunks)]
+#![feature(slice_as_chunks)]
+#![feature(portable_simd)]
 
 mod algorithms;
 mod bgworker;
--- a/src/prelude/distance.rs
+++ b/src/prelude/distance.rs
@@ -1,8 +1,8 @@
 use crate::prelude::*;
-use std::simd::f32x4;
-use std::simd::SimdFloat;
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
+use std::simd::f32x4;
+use std::simd::SimdFloat;
 
 const IS_VECTORIZARION_ENABLED: bool = is_vectorization_enabled();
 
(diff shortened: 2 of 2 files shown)


usamoi:
Compiler vectorization works on current codebase. Does it make sense to add a portable simd implementation?

Nicola115:
> 

Not sure is compiler vectorization enough for current use cases?

usamoi:
It generates compact AVX2 instructions on my platform. Not tested on a avx512 target or arm target so far.

Nicola115:
I see. we could leave the pr for the future and did further investigation first.

CyanPineapple:
The compiler struggles with auto-vectorize most of the time due to many constraints. 
Let's take `cosine distance` for example:
I put the scalar version into [godbolt](https://godbolt.org/z/j4KhcW75f) (with -O enabled). The asm between line16 - line22 shows it did scalar op (`ss`) rather than packed op (`ps`). The main reason is bound checking for `rhs` (`.LBB0_7`) in every iteration. Additionally, `lhs.len()` can't be calculated up front, it also constraints auto-vectorization. 
For performance, manual vectorization (portable-simd) might be necessary, but it's not an immediate necessity.
We can first optimize the current simd implementation. For example, can we substitute "a += b * c" with packed fma? 
Let's try

usamoi:
@CyanPineapple

There are three reasons why your code is not vectorized.

1. You need perform a bound check manually, eg. `assert_eq!(lhs.len(), rhs.len())`.
2. You need use `f*_fast` intrinsics to allow optimizations based on algebraic rules.
3. You need add `-Ctarget-cpu=x86-64-v3` to enable AVX2.

https://godbolt.org/z/hGWbTnP4r

You will see the main part of the code is compiled to compact AVX2 instructions.

CyanPineapple:
Thanks. `f*_fast` did help. Applying the optimization to scalar version resulted in a 10x speedup in my benchmark.

PR #51, opened 2023-08-17
fix: use x86-64-v3 as target cpu

usamoi:
Currently releases built by GitHub runners cannot run on machines without AVX-512 instructions. "x86-64-v3" relaxs the restriction for target instruction sets.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [build]
-rustflags = ["-Ctarget-cpu=native"]
+rustflags = ["-Ctarget-cpu=x86-64-v3"]
 rustdocflags = ["--document-private-items"]
 
 [target.'cfg(target_os="macos")']


VoVAllen (approved):

TIAcode:
```
cargo:warning=./src/f16.c:84:15: error: always_inline function '_mm512_set1_ps' requires target feature 'avx512f', but would be inlined into function 'v_f16_cosine_v4' that is compiled without support for 'avx512f'
```
The commit that fixed this a year ago isn't included in the sources anymore. Compiling this without AVX-512 runs into a ton of these errors. Would it still be possible to compile without it?

usamoi:
> ```
> cargo:warning=./src/f16.c:84:15: error: always_inline function '_mm512_set1_ps' requires target feature 'avx512f', but would be inlined into function 'v_f16_cosine_v4' that is compiled without support for 'avx512f'
> ```
> 
> The commit that fixed this a year ago isn't included in the sources anymore. Compiling this without AVX-512 runs into a ton of these errors. Would it still be possible to compile without it?

Can you explain what problems do you want to solve? `v_f16_cosine_v4` is decorated by `__attribute__((target("arch=x86-64-v4,avx512fp16")))` so there should not be this warning.

PR #81, opened 2023-09-25
feat: Quantization for HNSW

usamoi:
Changelog:

1. Mark `target-cpu=x86-64-v3` only for x86_64.
2. Quantization for HNSW.
3. Allow setting compression ratio for product quantization.
4. Use residuals for IVF.


VoVAllen (approved):
LGTM! Fantastic jobs!

PR #88, opened 2023-10-08
refactor: use mmap to replace socket for ipc

### LINKED ISSUE #70: optimization: Use mmap instead of socket for intra process communication
Currently we used socket for ipc, but mmap can be faster. As we only need single producer, single consumer pattern, we can make it even faster than https://github.com/cloudflare/mmap-sync

silver-ymz:
WIP
Close #70

Current problem:
`mmap-sync` doesn't provide a way to check if the writer has completed a write operation. Performing a read operation before starting a write can even cause a panic. Maybe we should try using self-managed mmap and condvar.


src/ipc/mmap/synchronizer.rs
```diff
@@ -0,0 +1,69 @@
...
+            reader_path.push_str("_server");
+        }
+        let mut reader = Synchronizer::new(reader_path.as_ref());
+        let writer = Synchronizer::new(writer_path.as_ref());
+
+        // The reader need to write something first to initialize the mmap file and state.
+        reader
+            .write(&[0u8; 1], Duration::from_secs(1))
+            .expect("Failed to write.");
+        unsafe { reader.read::<[u8; 1]>(false) }.expect("Failed to read.");
```
BeautyyuYanli, line 40: I think the writing is done if no exception raised? Because the mmap-sync is sync

silver-ymz: > I think the writing is done if no exception raised?

Yes. But the other side don't know when the writing is done.

BeautyyuYanli: I think the reader need to spawn a spin-loop thread to actively read from mmap to a buffer, and the ChannelTrait read is designed to read from the buffer when needed.

silver-ymz: > Is this the common pattern for stream based interface? For socket interface, receiver also don't know when the write operation is done by the writer?

For socket interface, the reader will be blocked until the write operation is done by the writer. But `mmap-sync` won't block anything. For read operations other than the first read, we can use spin-loop to block it mannually. But for the first time, it will directly panic.

> I think the reader need to spawn a spin-loop thread to actively read from mmap to a buffer

Using the way, we will give up the advantage of zero-copy with mmap.

BeautyyuYanli: > For read operations other than the first read, we can use spin-loop to block it mannually.

Since the mmap is the only IPC data link, I dont think there are any method for writer to "notify" the receiver that the writting is done.

> Using the way, we will give up the advantage of zero-copy with mmap.

The writer doesn't check if the sent data is consumed, so the next data will overwrite the previous one and causes "lossing packet"

silver-ymz: > Since the mmap is the only IPC data link, I dont think there are any method for writer to "notify" the receiver that the writting is done.

We can store extra state data in an additional mmap file. Because the memory is shared between 2 processes, all sync primitives can be used, like locks, etc.

> The writer doesn't check if the sent data is consumed, so the next data will overwrite the previous one and causes "lossing packet".

We can use write data in different place, also it's possible to check if the sent data is consumed. `mmap-sync` use [`RCU`](https://www.kernel.org/doc/html/next/RCU/whatisRCU.html) to implement similar function.

VoVAllen: Thanks for your explanation. @usamoi will implement this feature, and we're single producer, single consumer so it can be much easier. 
Spin-lock is not a good solution since it consumes lots of cpu resources.

VoVAllen: Seems https://github.com/servo/ipc-channel can be a good [baseline](https://github.com/dunnock/ipc-bench/tree/master)

PR #90, opened 2023-10-12
feat: add GUC option to manually control filter mode

### LINKED ISSUE #65: feat: GUC option to manually control using prefilter, postfilter, or skip index
User can `set vectors.filter.mode='skip/prefilter/postfilter'` to control how to use the vector index with filtering conditions

And add a function like https://stackoverflow.com/questions/49941564/change-guc-parameter-before-subquery-postgresql as shortcut

silver-ymz:
close #65

## test
When using skip mode,
```
explain select train.id from train order by 'a long vector' <=> train.embedding limit 32;
 Limit  (cost=490.54..499.02 rows=32 width=12)
   ->  Result  (cost=490.54..1815.81 rows=5001 width=12)
         ->  Sort  (cost=490.54..503.05 rows=5001 width=8)
               Sort Key: (('a long vector'::vector <=> embedding))
               ->  Seq Scan on train  (cost=0.00..340.51 rows=5001 width=8)
```

When using prefilter/postfilter, 
```
explain select train.id from train order by 'a long vector' <=> train.embedding limit 32;
 Limit  (cost=0.00..10.20 rows=32 width=12)
   ->  Index Scan using train_embedding_idx on train  (cost=0.00..1593.76 rows=5001 width=12)
         Order By: (embedding <=> 'a long vector'::vector)
```

And when using postfilter, it will output "Fallback to post-filter." like https://github.com/tensorchord/pgvecto.rs/blob/d96a11419a704388cd34545959d4f3962ac206e8/src/postgres/index_scan.rs#L162

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,9 @@ SET vectors.k = 32;
 SET LOCAL vectors.k = 32;
 ```
 
+If you want to disable vector indexing or prefilter, we also offer some GUC options:
+- `vectors.enable_vector_index`: Enable or disable the vector index. Default value is `on`.
+- `vectors.enable_prefilter`: Enable or disable the prefilter. Default value is `on`.
 
 ## Limitations
 
--- a/src/postgres/gucs.rs
+++ b/src/postgres/gucs.rs
@@ -1,4 +1,4 @@
-use pgrx::{GucContext, GucFlags, GucRegistry, GucSetting, PostgresGucEnum};
+use pgrx::{GucContext, GucFlags, GucRegistry, GucSetting};
 use std::ffi::CStr;
 
 pub static OPENAI_API_KEY_GUC: GucSetting<Option<&'static CStr>> =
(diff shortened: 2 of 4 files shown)


usamoi:
`enable_prefilter` might be a better name since it keeps consistency with https://www.postgresql.org/docs/current/runtime-config-query.html.

silver-ymz:
> enable_prefilter might be a better name since it keeps consistency with https://www.postgresql.org/docs/current/runtime-config-query.html.

`enable_prefilter` will be a boolean option. Does this mean we don't distinguish `prefilter` and `postfilter`?

usamoi:
> > enable_prefilter might be a better name since it keeps consistency with https://www.postgresql.org/docs/current/runtime-config-query.html.
> 
> `enable_prefilter` will be a boolean option. Does this mean we don't distinguish `prefilter` and `postfilter`?

If somebody needs, we can add a `force_prefilter`.

VoVAllen:
@usamoi How can user force post filter?

VoVAllen:
We can have `enable_filter` and `enable_prefilter` for now. What do you think?

VoVAllen:
Please also update corresponding the README

usamoi:
> We can have `enable_filter` and `enable_prefilter` for now. What do you think?

If there is a `where` and we do not really filter, won't it return incorrect data?

VoVAllen:
@usamoi You're right. It should be `enable_vector_index` and `enable_prefilter` ?

usamoi:
> @usamoi You're right. It should be `enable_vector_index` and `enable_prefilter` ?

Yes.

VoVAllen (approved):

usamoi (approved):

PR #92, opened 2023-10-16
ci: run ci when updating tests dir

silver-ymz:
Update:
- complete e2e test ci
- add github cache to avoid installing `pgrx` and `sqllogictest`.
- add `sccache` to share cache for each compilation of rust code. We don't use github cache directly because incremental compilation of rust will make the cache dir becoming larger and lager.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,14 +37,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Prepare
-      run: |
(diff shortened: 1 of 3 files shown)


VoVAllen:
You can try https://github.com/ankane/setup-postgres as what pgvector if it can make everything simpler. Also please consider merge the build and test sections, so that it won't build twice

silver-ymz:
> Also please consider merge the build and test sections, so that it won't build twice

Build will use debug mode, test will use release mode. Otherwise building the index in e2e test will take a long time. So, we can't avoid build twice.

In the future, we can use github cache to avoid installing and building dependencies. This will save considerable time.

VoVAllen:
@all-contributors please add @silver-ymz for code contribution, tests and infra

PR #95, opened 2023-10-17
feat: add Python bindings by SQLAlchemy

BeautyyuYanli:
Signed-off-by: Aurutus <emslhy@hotmail.com>

Continue the work of #75. Using SQLAlchemy to make python binding

Issue #58, opened 2023-08-31
feat: Implement python binding


gaocegege:
Maybe like this: https://github.com/pgvector/pgvector-python

AuruTus:
hi master @gaocegege , I'm interested in this, but can you provide more details? I have some thoughts but am still a bit confused.

AuruTus:
@VoVAllen  hi master, sorry for the late reply 😅. A bit busy with new semester enrolment. I'm alr in the discord channel, but I didn't found your name to at. Could you at me later when you are free?

BeautyyuYanli:
I'm currently working on this and the following is considered to be implemented. *The list may update frequently.*

Low-level python binding is to add support of vector-related Python structures for the Python postgres libraries.

---
From the view of libraries:
- [x] Support [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy) (ref: https://github.com/pgvector/pgvector-python#sqlalchemy)
- [ ] Support [Psycopg 3](https://github.com/psycopg/psycopg) (ref: https://github.com/pgvector/pgvector-python#psycopg-3)
- [ ] Support [Psycopg 2](https://github.com/psycopg/psycopg2) (ref: https://github.com/pgvector/pgvector-python#psycopg-2)
- [ ] Support [SQLModel](https://github.com/tiangolo/sqlmodel) (ref: https://github.com/pgvector/pgvector-python#sqlmodel)
- [ ] Support [asyncpg](https://github.com/MagicStack/asyncpg) (ref: https://github.com/pgvector/pgvector-python#asyncpg)
- [ ] Support [Peewee](https://github.com/coleifer/peewee) (ref: https://github.com/pgvector/pgvector-python#peewee)
---
From the view of Python structures:
- [ ] Read/Write 1-d np.array
	- [ ] Considering #86 , elements of vector may be u8
- [ ] Index options by Python dict/list structure, instead of TOML string

PR #102, opened 2023-10-24
feat: add Python bindings by psycopg 3

BeautyyuYanli:
relate to #58 
based on #95

### DIFF
--- a/bindings/python/src/pgvecto_rs/psycopg/__init__.py
+++ b/bindings/python/src/pgvecto_rs/psycopg/__init__.py
@@ -4,23 +4,18 @@ from psycopg.pq import Format
 from psycopg.types import TypeInfo
 from numpy import ndarray
 from ..utils.serializer import from_db_str, to_db_str
-from typing import List, Union
 
-__all__ = ['register_vector']
+__all__ = ["register_vector"]
 
 
 class VectorDumper(Dumper):
-
     format = Format.TEXT
 
     def dump(self, obj):
         return to_db_str(obj).encode("utf8")
 
 
-
-
 class VectorLoader(Loader):
-
     format = Format.TEXT
 
     def load(self, data):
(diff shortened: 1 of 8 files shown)


BeautyyuYanli:
@kemingy

bindings/python/tests/test_sqlalchemy.py
```diff
@@ -69,11 +69,11 @@ def test_create_index(session: Session, index_name: str, index_setting: str):
         index_name,
         Document.embedding,
         postgresql_using="vectors",
-        postgresql_with={"options": index_setting},
+        postgresql_with={"options": f"$${index_setting}$$"},
         postgresql_ops={"embedding": "l2_ops"},
     )
     index.create(session.bind)
-    session.rollback()
+    session.commit()
```
kemingy, line 76: Why does this action need to be committed while others can be rollback?

bindings/python/src/pgvecto_rs/psycopg/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    format = Format.TEXT
+
+    def load(self, data):
+        if isinstance(data, memoryview):
+            data = bytes(data)
+        return from_db_str(data.decode("utf8"))
+
+
+def register_vector(context: Connection):
+    info = TypeInfo.fetch(context, "vector")
```
kemingy, line 28: Are we using "vector" or "vectors"?

bindings/python/src/pgvecto_rs/psycopg/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    register_vector_info(context, info)
+
+
+def register_vector_info(context: Connection, info: TypeInfo):
+    if info is None:
+        raise ProgrammingError("vector type not found in the database")
+    info.register(context)
+
+    # add oid to anonymous class for set_types
+    text_dumper = type("", (VectorDumper,), {"oid": info.oid})
```
kemingy, line 43: Does it have to be an anonymous class?

bindings/python/src/pgvecto_rs/psycopg/__init__.py
BeautyyuYanli: The extension name is "vectors" and the type name is "vector"

bindings/python/tests/test_sqlalchemy.py
BeautyyuYanli: I guess some problems emerge when the index exists. I'm still trying.

bindings/python/src/pgvecto_rs/psycopg/__init__.py
BeautyyuYanli: Just a temporary wrapper class, and the value of info.oid is dynamically. Using `type(...)` is enough and clear.

kemingy: When will the `oid` be used?

Why is it an anonymous class instead of a normal one? I don't think a dynamic value will have to make the class anonymous.

BeautyyuYanli: I tried a normal defined version and it worked.

The oid will be used by the underlaying psycopg things.

kemingy (approved):

PR #110, opened 2023-10-31
fix: disable cache for cargo binary

usamoi:
Fix https://github.com/tensorchord/pgvecto.rs/actions/runs/6701671369/job/18209452281#step:5:143 .

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -42,7 +42,6 @@ jobs:
     - uses: actions/cache/restore@v3
       with:
         path: |
-          ~/.cargo/bin/
           ~/.cargo/registry/index/
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
@@ -72,7 +71,6 @@ jobs:
     - uses: actions/cache/restore@v3
       with:
         path: |
-          ~/.cargo/bin/
           ~/.cargo/registry/index/
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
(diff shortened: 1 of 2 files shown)


usamoi:
@silver-ymz Can you review this PR?

silver-ymz:
> Can you review this PR?

Yes, this change is right.

.github/workflows/check.yml
```diff
@@ -118,13 +115,12 @@ jobs:
         wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
         sudo apt-get update
         sudo apt-get -y install libpq-dev postgresql-${{ matrix.version }} postgresql-server-dev-${{ matrix.version }}
-        cargo install cargo-pgrx --git https://github.com/tensorchord/pgrx.git --rev $(cat Cargo.toml | grep "pgrx =" | awk -F'rev = "' '{print $2}' | cut -d'"' -f1) || true
+        cargo install cargo-pgrx --git https://github.com/tensorchord/pgrx.git --rev $(cat Cargo.toml | grep "pgrx =" | awk -F'rev = "' '{print $2}' | cut -d'"' -f1)
```
silver-ymz, line 118:
```suggestion
        (cargo install cargo-pgrx --git https://github.com/tensorchord/pgrx.git --rev $(cat Cargo.toml | grep "pgrx =" | awk -F'rev = "' '{print $2}' | cut -d'"' -f1)) || true
```

After researching, I found that shell script must add bracket in previous command like `(command) || true` to ignore the error code.

If we want to re-enable the cache for binary, we can add bracket for each commands.

usamoi: Why would cargo return a status which is not 0?

silver-ymz: Because when the cache was successfully restored, the `cargo-pgrx` had already been downloaded.

usamoi: It seems that the broken action is caused by incorrect .so library version. If we re-enable cache for binaries, we may need to lock versions of system C/C++ dependencies. Do you have ideas about it?

silver-ymz: Since that, disabling cache for binaries may be the better way.

silver-ymz (approved):

PR #118, opened 2023-11-09
chore: update the extention description

kemingy:


usamoi:
This PR is breaking. We must use `tensorchord.vector(1000)` instead of `vector(1000)` and `USING tensorchord.vectors` instead of `USING vectors`.

kemingy:
> This PR is breaking. We must use `tensorchord.vector(1000)` instead of `vector(1000)` and `USING tensorchord.vectors` instead of `USING vectors`. Operators do not work now.

I think it's better to have a separate schema instead of the `public` one since there are multiple extensions called "vector".

As for the breaking, we only need to update the doc.

usamoi:
Not everything can be put into a separate schema. Schema-qualified operators are hard to use.[^0]

[^0]: https://dba.stackexchange.com/questions/191695/is-it-possible-to-schema-qualify-an-operator

VoVAllen:
Use a seperate PR for dco https://github.com/tensorchord/pgvecto.rs/pull/121

usamoi:
If users need, they can use `CREATE EXTESNION vectors WITH SCHEMA tensorchord`.

kemingy:
This is how paradedb does:
- https://github.com/paradedb/paradedb/blob/2b3919a20edd5bc40724e26a56b84538e70d66ec/scripts/entrypoint.sh#L75

Issue #108, opened 2023-10-29
bug: ERROR:  operator is not unique: unknown <-> unknown


BeautyyuYanli:
```
❯ psql -h localhost -U postgres
Password for user postgres: 
psql (15.4)
Type "help" for help.

postgres=# SELECT '[1, 2, 3]' <-> '[3, 2, 1]';
ERROR:  operator is not unique: unknown <-> unknown
LINE 1: SELECT '[1, 2, 3]' <-> '[3, 2, 1]';
                           ^
HINT:  Could not choose a best candidate operator. You might need to add explicit type casts.
```

silver-ymz:
The default type of '' is unknown in SQL. Seems you need to add explicit type casts, e.x. `SELECT '[1,2,3]'::vector <-> '[1,2,3]'::vector;`

BeautyyuYanli:
`<#>` and `<=>` is ok

Need to update readme

PR #122, opened 2023-11-09
chore: replace '<->' with '<*>' for L2 operator

usamoi:
`<->` is overloaded by PostgreSQL's text search facility. [^1]

We should use another operator name to escape potential problems.

related: #108 

[^1]: https://www.postgresql.org/docs/current/functions-textsearch.html


usamoi:
Python Check will fail because it pulls latest docker image instead of building source from repo.

VoVAllen:
I've raised the similar issue in the pgvector repo. Let's see what's their responses.

PR #123, opened 2023-11-09
feat: Add high-level API for Python

BeautyyuYanli:


BeautyyuYanli:
I may add docs in another PR @kemingy

bindings/python/src/pgvecto_rs/highapi/client.py
```diff
@@ -0,0 +1,199 @@
...
+            text: Mapped[str] = mapped_column(String)
+            meta: Mapped[dict] = mapped_column(postgresql.JSONB)
+            embedding: Mapped[ndarray] = mapped_column(Vector(dimension))
+
+        self._engine = create_engine(db_url)
+        with Session(self._engine) as session:
+            session.execute(text("CREATE EXTENSION IF NOT EXISTS vectors"))
+            session.commit()
+        self._table = _Table
+        self._table.__table__.create(self._engine, checkfirst=not new_table)  # type: ignore
```
kemingy, line 69: What's the purpose of this?

bindings/python/src/pgvecto_rs/highapi/client.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+
+        class _Table(RecordORM):
+            __tablename__ = f"table_{table_name}"
+            id: Mapped[UUID] = mapped_column(
+                postgresql.UUID(as_uuid=True), primary_key=True
+            )
+            text: Mapped[str] = mapped_column(String)
+            meta: Mapped[dict] = mapped_column(postgresql.JSONB)
+            embedding: Mapped[ndarray] = mapped_column(Vector(dimension))
```
kemingy, line 62: I don't get it. Why do we need to inherit `RecordORM` then override all the attributes?

bindings/python/src/pgvecto_rs/highapi/client.py
```diff
@@ -0,0 +1,199 @@
...
+    ):
+        client = cls(db_url, table_name, dimension, embedder, True)
+        meta = meta or {}
+        try:
+            for i in range(len(texts)):
+                client.add_text(texts[i], meta)
+        except Exception as e:
+            client.drop()
+            raise e
+        return client
```
kemingy, line 118: Why do we need to construct the class from text/records?

bindings/python/src/pgvecto_rs/highapi/client.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # ================ Query ================
+
+    def search(
+        self,
+        embedding: Union[ndarray, List[float]],
+        distance_op: Literal["<->", "<=>", "<#>"] = "<->",
+        limit: int = 10,
+        filter: Optional[FilterFunc] = None,
+        order_by_dis: bool = True,
```
kemingy, line 149: I think this is 100% True. Are there any requirements to return the results without ordering?

bindings/python/src/pgvecto_rs/highapi/record.py
```diff
@@ -0,0 +1,49 @@
...
+from numpy import array, float32, ndarray
+from sqlalchemy.orm import DeclarativeBase, Mapped
+
+
+class RecordORM(DeclarativeBase):
+    __tablename__: str
+    id: Mapped[UUID]
+    text: Mapped[str]
+    meta: Mapped[dict]
+    embedding: Mapped[ndarray]
```
kemingy, line 13: Why is it a numpy.ndarray instead of List[float]?

BeautyyuYanli: The underlying driver is designed to use np.ndarry first (following pgvector)

bindings/python/src/pgvecto_rs/highapi/client.py
BeautyyuYanli: It's just how the SQLalchemy works...

bindings/python/src/pgvecto_rs/highapi/record.py
```diff
@@ -0,0 +1,49 @@
...
+[meta]     : {self.meta}
+[embedding]: {self.embedding}
+========== End of Record ========="""
+
+    @classmethod
+    def from_orm(cls, orm: RecordORM):
+        return cls(orm.id, orm.text, orm.meta, orm.embedding)
+
+    @classmethod
+    def from_text(
```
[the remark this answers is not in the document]
BeautyyuYanli: Record is designed to be exposed to users, so I don't want to mess it with SQLAlchemy things...

bindings/python/src/pgvecto_rs/highapi/client.py
BeautyyuYanli: The init function will connect to an existing table or create an empty new one if not exists.

bindings/python/src/pgvecto_rs/sdk/client.py
```diff
@@ -0,0 +1,130 @@
...
+            session.commit()
+        self._table = _Table
+        self._table.__table__.create(self._engine)
+        self.dimension = dimension
+
+    def upsert(self, records: List[Record]) -> None:
+        with Session(self._engine) as session:
+            for record in records:
+                session.execute(
+                    insert(self._table).values(
```
kemingy, line 64: What does "upsert" here mean? Will the ORM handle the id collision and update the record?

bindings/python/src/pgvecto_rs/sdk/record.py
```diff
@@ -0,0 +1,45 @@
...
+[embedding]: {self.embedding}
+========== End of Record ========="""
+
+    @classmethod
+    def from_orm(cls, orm: RecordORM):
+        return cls(orm.id, orm.document, orm.embedding)
+
+    @classmethod
+    def from_text(
+        cls, document: Optional[dict], embedding: Union[ndarray, List[float]]
```
kemingy, line 41: Using the dict parameter is a bad idea. It's only used when you have to. BTW, `from_text` but has no `text` required is also a bad interface.

Filtering is an add-on feature, not an everyone-must-use feature. Users have to create a dict even when they don't have any metadata for filtering.

bindings/python/src/pgvecto_rs/sdk/filters.py
```diff
@@ -0,0 +1,21 @@
...
+Filter = Type[Callable[[RecordORMType], ColumnElement[bool]]]
+FilterInput = RecordORMType
+FilterOutput = ColumnElement[bool]
+
+
+def document_contains(document_contains: dict) -> Filter:
+    return lambda r: r.document.contains(document_contains)
+
+
+__all__ = [
```
kemingy, line 16: This is only required when you need to expose them. Usually, in the `__init__.py` file.

bindings/python/src/pgvecto_rs/sdk/client.py
```diff
@@ -0,0 +1,130 @@
...
+            dimension (int): dimension of the embeddings.
+        """
+
+        class _Table(RecordORM):
+            __tablename__ = f"collection_{collection_name}"
+            id: Mapped[UUID] = mapped_column(
+                postgresql.UUID(as_uuid=True), primary_key=True
+            )
+            document: Mapped[dict] = mapped_column(postgresql.JSONB)
+            embedding: Mapped[ndarray] = mapped_column(Vector(dimension))
```
kemingy, line 50: Is it necessary to redeclare these attributes?

bindings/python/src/pgvecto_rs/sdk/record.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class RecordORM(DeclarativeBase):
+    __tablename__: str
+    id: Mapped[UUID]
+    document: Mapped[dict]
+    embedding: Mapped[ndarray]
+
+
+RecordORMType = Type[RecordORM]
```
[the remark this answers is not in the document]
BeautyyuYanli: Record is designed to be exposed to users, so I don't want to mess it with SQLAlchemy things...

bindings/python/src/pgvecto_rs/sdk/client.py
BeautyyuYanli: Yes, it's just how the SQLalchemy works...

bindings/python/examples/sqlalchemy_example.py
```diff
@@ -1,11 +1,10 @@
 import os
 
 import numpy as np
+from pgvecto_rs.sqlalchemy import Vector
```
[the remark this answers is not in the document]
BeautyyuYanli: I'm using the default config and latest ruff, but the behavior do have some differences with isort.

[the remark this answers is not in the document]
kemingy: Seems you will need to add "pgvecto_rs" to https://docs.astral.sh/ruff/settings/#isort-known-first-party. It should be able to detect that, but I don't know why it cannot.

bindings/python/pyproject.toml
```diff
@@ -1,40 +1,53 @@
...
+"psycopg[binary]>=3.1.12",
 ]
 
 [tool.pdm.dev-dependencies]
-dev = [
-    "SQLAlchemy>=2",
-    "pytest>=7.4",
-    "psycopg[binary]>=3.1.12",
+lint = [
+    "ruff>=0.1.1",
```
kemingy, line 36: align with the lock file

bindings/python/src/pgvecto_rs/sdk/record.py
```diff
@@ -0,0 +1,49 @@
...
+[embedding]: {self.embedding}
+========== End of Record ========="""
+
+    @classmethod
+    def from_orm(cls, orm: RecordORM):
+        return cls(orm.id, orm.text, orm.meta, orm.embedding)
+
+    @classmethod
+    def from_text(
+        cls, text: str, meta: Optional[dict], embedding: Union[ndarray, List[float]]
```
kemingy, line 45: If `meta` is optional, should it have a default value? Or do users have to pass a None since it's required?

BeautyyuYanli: Users have to pass a None

kemingy: It's weird to explicitly pass a `None`.

BeautyyuYanli: Modified

kemingy (changes requested):

kemingy (approved):

PR #124, opened 2023-11-09
fix: disallow  type `vector(0)`

usamoi:
Disallow type `vector(0)` so that we only accept `vector`, `vector(1)`, ... `vector(65535)`.


VoVAllen:
Is it `vectors(0)` or `vector(0)`? And will it raise error for `vector(0)`

usamoi:
`vector(0)`. Yes, it will raise.

PR #133, opened 2023-11-13
feat: mmap for ipc

### LINKED ISSUE #70: optimization: Use mmap instead of socket for intra process communication

usamoi:
Use mmap instead of unix domain socket for session-bgworker ipc.

The feature is enabled by default. Set `vectors.transport` to `unix` to use unix domain socket for ipc instead.

Closes #70.


usamoi:
We should bump a version after merging this PR. This PR prevents any extension loading excluding loading by `preload_shared_libraries` and gives a nice message. I think it will stop many issues reporting `IpcError::Closed`.

VoVAllen (dismissed):
What's the error message for unloaded "preload_shared_libraries"? Please give simple fix instructions to users in the error message

usamoi:
> What's the error message for unloaded "preload_shared_libraries"? Please give simple fix instructions to users in the error message

It's `pgvecto.rs must be loaded via shared_preload_libraries`. I think users will read `docs/INSTALL.md` if they do not make it work?

usamoi:
Current error message:
```
ERROR:  pgvecto.rs: pgvecto.rs must be loaded via shared_preload_libraries.
ADVICE: If you encounter this error for your first use of pgvecto.rs, please read `https://github.com/tensorchord/pgvecto.rs/blob/main/docs/install.md`. You should edit `shared_preload_libraries` in `postgresql.conf` to include `vectors.so`, or simply run the command `psql -U postgres -c 'ALTER SYSTEM SET shared_preload_libraries = "vectors.so"'`.
```

PR #138, opened 2023-11-16
feat: add more ruff rules

BeautyyuYanli:


BeautyyuYanli:
@kemingy mutable data as deafult arg is covered by rule B

bindings/python/pyproject.toml
```diff
@@ -40,19 +40,34 @@ sqlalchemy = [
...
+  "SIM", #https://docs.astral.sh/ruff/rules/#flake8-simplify-sim
+  "N", #https://docs.astral.sh/ruff/rules/#pep8-naming-n
+  "PT", #https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt
+  "TRY", #https://docs.astral.sh/ruff/rules/#tryceratops-try
+  "FLY", #https://docs.astral.sh/ruff/rules/#flynt-fly
+  "PL", #https://docs.astral.sh/ruff/rules/#pylint-pl
+  "NPY", #https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy
+  "RUF", #https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf
+]
+ignore = ["S101", "E731", "E501"]
```
kemingy, line 69: Do we have any `assert` for the runtime code?

bindings/python/src/pgvecto_rs/errors.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+
+
+class VectorError(ValueError):
```
kemingy, line 4: Might need to make the name more specific to this library name.

bindings/python/pyproject.toml
BeautyyuYanli: Not in src but in tests

kemingy: It should be allowed to use `assert` in tests. Seems it's a bug in ruff https://github.com/astral-sh/ruff/issues/4368.

kemingy (approved):

PR #139, opened 2023-11-16
ci: add release for arm64, pg14, pg16

silver-ymz:


.github/workflows/release.yml
```diff
@@ -100,7 +100,8 @@ jobs:
   release:
     strategy:
       matrix:
-        version: [15]
+        arch: [amd64, arm64]
+        version: [14, 15, 16]
```
jrasm91, line 104: Might be worth looking at [this](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations).

I think this would also work:
```yml
      matrix:
        arch: [amd64, arm64]
        version: [14, 15, 16]
        include:
          - arch: amd64
            full_arch: x86_64
          - arch: arm64
            full_arch: aarch64
```
Then you could delete the step to set the env and just use `${{ matrix.full_arch }}`

VoVAllen:
Have you tested this on your fork?

silver-ymz:
It can run successfully now.
To see https://github.com/silver-ymz/pgvecto.rs/actions/runs/6906504206/job/18791688511

.github/workflows/release.yml
silver-ymz: Thanks

VoVAllen:
https://github.com/all-contributors please add @silver-ymz for infra

VoVAllen:
@all-contributors please add @silver-ymz for infra

Issue #78, opened 2023-09-13
feat: progress information on index creation


VoVAllen:
Provide a function to show how the percentage of the tuples already processed

AuruTus:
I'm working on this. 😃

PR #155, opened 2023-11-21
feat: progress information on index creation

silver-ymz:
done #78

### DIFF
--- a/src/postgres/functions.rs
+++ b/src/postgres/functions.rs
@@ -1,1 +1,13 @@
+use super::hook_transaction::client;
+use crate::prelude::{Friendly, Id};
 
+#[pgrx::pg_extern(volatile, parallel_safe, strict)]
+fn vector_stat_tuples_done(oid: pgrx::pg_sys::Oid) -> i32 {
+    let id = Id::from_sys(oid);
+    let mut res = 0;
+    client(|mut rpc| {
+        res = rpc.stat(id).unwrap().friendly();
+        rpc
+    });
+    res.try_into().unwrap()
+}
--- a/src/sql/finalize.sql
+++ b/src/sql/finalize.sql
@@ -1,17 +1,31 @@
 CREATE CAST (real[] AS vector)
-	WITH FUNCTION cast_array_to_vector(real[], integer, boolean) AS IMPLICIT;
+    WITH FUNCTION cast_array_to_vector(real[], integer, boolean) AS IMPLICIT;
 
 CREATE CAST (vector AS real[])
-	WITH FUNCTION cast_vector_to_array(vector, integer, boolean) AS IMPLICIT;
+    WITH FUNCTION cast_vector_to_array(vector, integer, boolean) AS IMPLICIT;
 
 CREATE OPERATOR CLASS l2_ops
-	FOR TYPE vector USING vectors AS
-	OPERATOR 1 <-> (vector, vector) FOR ORDER BY float_ops;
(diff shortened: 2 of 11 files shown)


usamoi:
`waiting_secs` is a parameter for an unimplemented feature. If a growing segment do not grow for `waiting_secs` seconds, it will be sealed.

VoVAllen:
Can you also include the full index configuration? Such as the full json including quantization and other configuration? We can call the table as vector_index_info. And I think it's better to put it under a custom namespace, to avoid potential conflicts with user's table

VoVAllen:
Also is it a common practice to use a table instead of a function like `select xxx from info()`? What's pgvector's interface?

usamoi:
> Also is it a common practice to use a table instead of a function like `select xxx from info()`? What's pgvector's interface?

Yes, Postgres uses this way.

https://www.postgresql.org/docs/current/progress-reporting.html

silver-ymz:
> And I think it's better to put it under a custom namespace, to avoid potential conflicts with user's table

To avoid potential conflicts with user's table, maybe we can name it to `pg_vector_index_info`, because tables starting with "pg_" are usually considered as system reserved tables.

VoVAllen:
> > And I think it's better to put it under a custom namespace, to avoid potential conflicts with user's table
> 
> To avoid potential conflicts with user's table, maybe we can name it to `pg_vector_index_info`, because tables starting with "pg_" are usually considered as system reserved tables.

LGTM

silver-ymz:
> Can you also include the full index configuration?

done

```sql
select * from pg_vector_index_info;

 relid  | indexrelid | relname | indexrelname | idx_tuples | idx_tuples_done |                                                                                                                                                         idx_config                                                                                                                                                         
--------+------------+---------+--------------+------------+-----------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 389935 |     389940 | t       | t_val_idx    |     100000 |           60000 | {"vector":{"dimensions":3,"distance":"L2"},"segment":{"max_growing_segment_size":20000,"min_sealed_segment_size":1000,"max_sealed_segment_size":1000000},"optimizing":{"waiting_secs":60,"deleted_threshold":0.2,"optimizing_threads":2},"indexing":{"hnsw":{"m":12,"ef_construction":300,"quantization":{"trivial":{}}}}}
```

VoVAllen:
LGTM, can you also add the table name also?

silver-ymz:
> can you also add the table name also?

which table name?

usamoi:
waiting_secs and deleted_threshold is not public so far. Use serde(skip) to hide them.

silver-ymz:
> waiting_secs and deleted_threshold is not public so far. Use serde(skip) to hide them.

done

VoVAllen:
The table name of the index belongs to

silver-ymz:
> The table name of the index belongs to

`relname` is that.

VoVAllen:
pg also have reltype. I think it's a generic name. Let's just call it tablename and indexname

VoVAllen:
Also relid->tablerelid?

src/postgres/functions.rs
usamoi, line 1: The two functions cannot be marked with `immutable` or `parallel_safe`.

src/sql/finalize.sql
usamoi, line 1: We need to monitor more things. Even if indexed tuples equal to total tuples, the index might continue merging.

PR #158, opened 2023-11-22
Revert "feat: progress information on index creation"

usamoi:
Reverts tensorchord/pgvecto.rs#155


VoVAllen:
Found problem without sleeping second, temporarily revert

PR #165, opened 2023-11-22
feat: progress information on index creation 

silver-ymz:
Update since #155
- Add a futex in `Index`. wait when optimization is impossible, wake when refresh.
- Add `idx_indexing` in `pg_vector_index_info`
- Refactor `idx_tuples` to get from backend, because `pg_class` may be updated later than the backend and there will be cases where `idx_tuples` is less than `idx_tuples_done.


src/postgres/functions.rs
```diff
@@ -1 +1,47 @@
...
 
+#[pgrx::pg_extern(volatile, strict)]
+fn vector_stat_indexing(oid: pgrx::pg_sys::Oid) -> bool {
+    let id = Id::from_sys(oid);
+    let mut res = MaybeUninit::uninit();
+    client(|mut rpc| {
+        res.write(rpc.stat_indexing(id).unwrap().friendly());
+        rpc
+    });
+    unsafe { res.assume_init() }
```
usamoi, line 13: Do not use unsafe code here.

src/ipc/server.rs
```diff
@@ -92,6 +116,22 @@ pub enum RpcHandle {
         id: Id,
         x: Destory,
     },
+    StatIndexing {
```
usamoi, line 119: I think we need to monitor status of all segments. It's Okay to leave it for future.

src/index/mod.rs
```diff
@@ -400,10 +418,14 @@ impl IndexBackground {
             return;
         }
         while let Some(index) = self.index.upgrade() {
-            pool.install(|| {
-                optimizing::indexing::optimizing_indexing(index.clone());
-            });
-            std::thread::sleep(Duration::from_secs(60));
+            index.state.store(1, Relaxed);
+            let done = pool.install(|| optimizing::indexing::optimizing_indexing(index.clone()));
+            if done {
```
usamoi, line 423: it's possible to continuing creating indexes without new data coming

src/index/mod.rs
```diff
@@ -70,6 +73,9 @@ pub struct Index {
     delete: Arc<Delete>,
     protect: Mutex<IndexProtect>,
     view: ArcSwap<IndexView>,
+    // the first bit represents whether it is indexing
+    // the second bit represents whether there is new data
+    state: AtomicU32,
```
usamoi, line 78: Do not suggest an atomic here.

src/index/mod.rs
silver-ymz: However, the return value of `optimizing_indexing` will only be true if `segs_len < index.options.segment.min_sealed_segment_size as u64 || segs.len() < 3`. This depends on `index.protect`; if `index.protect` is not updated, the condition will always be true. And only `refresh` funtion will update `index.protect`.

usamoi: I'm not sure if `index.state.store(1, Relaxed);` can be reorder after `if`. It's really confusing.

src/postgres/functions.rs
silver-ymz: done in new commit

src/index/mod.rs
silver-ymz: update `index.state.store(1, Relaxed);` to `SeqCst` and others to corresponding `Release` and `Acquire`

src/index/mod.rs
silver-ymz: Is there any better solution?

usamoi: You can use a thread parker.

silver-ymz:
Update:
- refactor atomic to thread parker. But it seems that we can't directly know if it's indexing.
- remove `idx_indexing`, add `idx_sealed`, `idx_growing`

`pg_vector_index_info` example:
```
 tablerelid | indexrelid | tablename | indexname | idx_tuples | idx_tuples_done | idx_sealed |  idx_growing  |                                                                                                                                    idx_config                                                                                                                                    
------------+------------+-----------+-----------+------------+-----------------+------------+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
     390466 |     390471 | t         | t_val_idx |     200000 |          180000 | {140000}   | {20000,20000} | {"vector":{"dimensions":3,"distance":"L2"},"segment":{"max_growing_segment_size":20000,"min_sealed_segment_size":1000,"max_sealed_segment_size":1000000},"optimizing":{"optimizing_threads":2},"indexing":{"hnsw":{"m":12,"ef_construction":300,"quantization":{"trivial":{}}}}}
```

src/bgworker/mod.rs
```diff
@@ -117,6 +117,26 @@ fn session(worker: Arc<Worker>, mut handler: RpcHandler) -> Result<(), IpcError>
                 worker.call_destory(id);
                 handler = x.leave()?;
             }
+            RpcHandle::StatTuples { id, x } => {
```
usamoi, line 120: Can we merge `StatTuples`, `StatTuplesDone`, `StatSealed`, `StatGrowing` to one function? The info about write segment is useful too.

src/index/mod.rs
```diff
@@ -70,6 +71,7 @@ pub struct Index {
     delete: Arc<Delete>,
     protect: Mutex<IndexProtect>,
     view: ArcSwap<IndexView>,
+    optimize_unparker: OnceLock<Unparker>,
```
usamoi, line 74: You need to read the document and examples of parker. What you implemented before is just a parker.

Knowing if indexing is building is still needed.

silver-ymz: done in new commit

src/bgworker/mod.rs
silver-ymz: done in new commit

src/index/mod.rs
usamoi: Why OnceLock?

Why unpark is called by optimizing thread?

silver-ymz: > Why OnceLock?

The optimizing thread will be created after index is created. So we need `OnceLock` to lazy set the unparker.

> Why unpark is called by optimizing thread?

Unpark will be called by both refresh and optimizing thread. Unpark in optimizing thread is used to clean the exist token, otherwise the park operation will be unparked by refresh function called before starting optimizing.

silver-ymz: > Why unpark can be used to clean the existing token?

Because `unpark` won't remember previous token. No mater what previous state is, after doing `u.unpark(); p.park();`, the token will be cleared.

usamoi: Why clear the token?

silver-ymz:
```
refresh
start optimizing
done -> park
```

If events happen with this order, the park function will return immediately and it will only be parked in next loop.

silver-ymz: enter `optimizing_indexing` function

usamoi: Is not the token consumed by line 436?

silver-ymz: If refresh function is called again during optimizing, it won't be parked.

usamoi: I see. I do not think it is a problem. Compared to time-consuming building index, checking if an index is needed is not so important.

silver-ymz: Removed in new commit.

src/index/mod.rs
```diff
@@ -228,10 +242,23 @@ pub struct IndexView {
 }
 
 impl IndexView {
-    #[allow(dead_code)]
     pub fn len(&self) -> u32 {
+        self.sealed_len() + self.growing_len() + self.write_len()
```
usamoi, line 246: These functions should be merged too, otherwise inconsistency data might be seen.

usamoi: I think we should just send information as many as possible.

silver-ymz: https://github.com/silver-ymz/pgvecto.rs/blob/index-progress/src/bgworker/worker.rs#L132-L152

The functions are all called from the same view, the data will always be consistent.

usamoi: `write_len` may change.

silver-ymz: > write_len may change.

write is an `Arc` in `IndexView`. It won't be changed. 

In worker.rs:133, `ArcSwap::load_full` will make a copy for view arc.

silver-ymz: updated in new commit

usamoi (approved):

PR #174, opened 2023-11-28
fix: pg16 updates RelFileNode to RelFileLocator

silver-ymz:


silver-ymz:
> is it ready to be merged?

Yes

PR #177, opened 2023-12-03
feat: vbase optimization

### LINKED ISSUE #171: feat: Post-filter with VBASE

silver-ymz:
close #171

### DIFF
--- a/src/algorithms/hnsw.rs
+++ b/src/algorithms/hnsw.rs
@@ -54,6 +54,15 @@ impl Hnsw {
     pub fn search(&self, k: usize, vector: &[Scalar], filter: &mut impl Filter) -> Heap {
         search(&self.mmap, k, vector, filter)
     }
+
+    pub fn search_vbase<'index, 'vector, 'filter>(
+        &'index self,
+        range: usize,
+        vector: &'vector [Scalar],
+        filter: *mut (impl Filter + 'filter),
+    ) -> HnswIndexIter<'index, 'vector, 'filter> {
+        search_vbase(&self.mmap, range, vector, filter)
+    }
 }
 
 unsafe impl Send for Hnsw {}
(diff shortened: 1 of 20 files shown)


gaocegege:
Thanks for your contribution! :tada: :+1:

src/algorithms/hnsw.rs
```diff
@@ -509,6 +531,55 @@ pub fn local_search(
...
+            if !visited.check(v) {
+                continue;
+            }
+            visited.mark(v);
+            let v_dis = mmap.quantization.distance(mmap.d, vector, v);
+            if !results.check(v_dis) {
+                continue;
+            }
+            candidates.push(Reverse((v_dis, v)));
+            results.push(HeapElement {
```
usamoi, line 567: Will it lost data to be returned?

silver-ymz: Yes.

usamoi: The first stage points should be saved and sorted. In the second stage, we will return a better choice in first stage points and second stage heap.

usamoi: Not all, at most `range` points here.

silver-ymz: The line 578 saves the the first stage heap to the iterator. The current implementation has already chosen the first stage point.

usamoi: First stage heap contains at most `range` points. All points in first stage, including points that are not in first stage heap, should be returned.

silver-ymz: What do you mean by the points in the first stage but not in the first stage heap?

usamoi: All visited points. `results` will contain at most `range` points, and points are deleted by `heap::push` are lost.

src/algorithms/hnsw.rs
```diff
@@ -509,6 +531,61 @@ pub fn local_search(
...
+    assert!(range > 0);
+    let mut visited_guard = mmap.visited.fetch();
+    let mut visited = visited_guard.fetch();
+    let mut candidates = BinaryHeap::<Reverse<(Scalar, u32)>>::new();
+    let mut results = Heap::new(range);
+    let mut heap = BinaryHeap::<Reverse<HeapElement>>::new();
+    visited.mark(s);
+    let s_dis = mmap.quantization.distance(mmap.d, vector, s);
+    candidates.push(Reverse((s_dis, s)));
+    results.push(HeapElement {
```
usamoi, line 549: Push to `heap`.

A vec to be sorted is better here.

usamoi: Push `s` to heap.

`heap` need to be sorted.

Compare head of `heap` and `result` to return the next result .

silver-ymz: Fixed in new commit. Could you review again?

PR #203, opened 2023-12-19
fix: upgrade instructions

usamoi:
all older version of install scripts are uploaded, we need to write upgrade scripts later


gaocegege:
There are some conflicts in src/libs.rs, please resolve them.

VoVAllen (approved):

PR #206, opened 2023-12-24
feat: allow user to set vbase range

### LINKED ISSUE #201: feat: allow user to set vbase range

silver-ymz:
close #201


silver-ymz:
It's ready to review. @VoVAllen @usamoi

VoVAllen (commented):
Do we need to separate the `enable_vbase` and `vbase_range` settings? I prefer to seperate them, so user doesn't need to take care of the vbase range value in most scenarios. My suggestion is let default vbase_range to be 100 and separate these two options. 

@usamoi WDYT?

usamoi (approved):

PR #212, opened 2023-12-28
feat: add nprobe into guc

### LINKED ISSUE #194: feat: Set ivf.probes in runtime
Currently it's in the index options, should be defined as runtime variable

cutecutecat:
Fix #194 

# Design
It's a bad idea to straightly pass parameter `nprobe` to all indexes, as only `ivf` need it.
However, for `ipc` caller `am_scan`, it's transparent of index type in use, as it is parsed at callee `service`, so we could not construct an union before `am_scan`.

Thus, we create a dataclass `SearchGucs` which could handle runtime variables for both `sealed` and `growing` segments.
And `SealedSearchGucs` which contains `ivf_probe` and other future required runtime variables, it could handle all runtime variables for sealed indexes, thus approprate to pass to all indexes.

If there is other idea for the design, I would be happy to know.

# Feat
- Introduce `SearchGucs` for gucs accessed by `search` method
- Introduce `SealedSearchGucs` for gucs accessed by `sealed segments search` method
- `ivf` would handle `SealedSearchGucs.ivf_nprobe` while `HNSW` and `flat` not, leaving a private argument `_gucs`

### DIFF
--- a/crates/service/src/algorithms/ivf/ivf_naive.rs
+++ b/crates/service/src/algorithms/ivf/ivf_naive.rs
@@ -55,8 +55,14 @@ impl<S: G> IvfNaive<S> {
         self.mmap.raw.payload(i)
     }
 
-    pub fn search(&self, k: usize, vector: &[S::Scalar], filter: &mut impl Filter) -> Heap {
-        search(&self.mmap, k, vector, filter)
+    pub fn search(
+        &self,
+        k: usize,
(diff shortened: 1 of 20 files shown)


gaocegege:
PTAL @VoVAllen @usamoi

usamoi (approved):

gaocegege:
@all-contributors Please add @cutecutecat  for the code contribution.

PR #220, opened 2024-01-02
chore: fine-grained upgrade hint

### LINKED ISSUE #192: bug: Support REINDEX/DROP INDEX when breaking change
Related issue: https://github.com/tensorchord/pgvecto.rs/issues/191

usamoi:
closes #192

For the whole data directory and per-index directory, there will be a file named `metadata`. It's a JSON and there are two fields, `version` and `soft_version`.

`version` is the break-changing version. If the `version` is different with `.so` file. The bgworker does not work or the index does not work.
`soft_version` is the soft version, if the `soft_version` is equal or less than `.so` file. The bgworker/index will continue to work.


usamoi:
@VoVAllen Could you review it?

PR #222, opened 2024-01-02
feat: show index size in index info table

### LINKED ISSUE #185: feat: Show index size in index info table

kemingy:
- fix #185 

```log
vectors=# select tablerelid, idx_size from pg_vector_index_info ;
 tablerelid | idx_size                                         
      16873 | 4
(1 row)
```

### DIFF
--- a/crates/service/src/index/mod.rs
+++ b/crates/service/src/index/mod.rs
@@ -69,6 +69,7 @@ pub struct IndexStat {
     pub growing: Vec<u32>,
     pub write: u32,
     pub options: IndexOptions,
+    pub size: u64,
 }
 
 pub struct Index<S: G> {
@@ -239,6 +240,8 @@ impl<S: G> Index<S> {
             growing: view.growing.values().map(|x| x.len()).collect(),
             write: view.write.as_ref().map(|(_, x)| x.len()).unwrap_or(0),
             options: self.options().clone(),
+            size: view.growing.values().map(|x| x.size() as u64).sum::<u64>()
+                + view.sealed.values().map(|x| x.size() as u64).sum::<u64>(),
         }
     }
 }
--- a/crates/service/src/index/mod.rs
+++ b/crates/service/src/index/mod.rs
@@ -63,13 +63,23 @@ pub struct IndexOptions {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SegmentSizeInfo {
+    pub id: Uuid,
+    #[serde(rename = "type")]
+    pub typ: String,
+    pub length: usize,
+    pub size: u64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 pub struct IndexStat {
     pub indexing: bool,
     pub sealed: Vec<u32>,
     pub growing: Vec<u32>,
     pub write: u32,
     pub options: IndexOptions,
-    pub size: u64,
+    pub size: Vec<SegmentSizeInfo>,
 }
 
 pub struct Index<S: G> {
@@ -240,8 +250,12 @@ impl<S: G> Index<S> {
             growing: view.growing.values().map(|x| x.len()).collect(),
             write: view.write.as_ref().map(|(_, x)| x.len()).unwrap_or(0),
             options: self.options().clone(),
-            size: view.growing.values().map(|x| x.size()).sum::<u64>()
-                + view.sealed.values().map(|x| x.size()).sum::<u64>(),
+            size: view
+                .sealed
+                .values()
+                .map(|x| x.size())
+                .chain(view.growing.values().map(|x| x.size()))
+                .collect(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


crates/service/src/index/segments/sealed.rs
```diff
@@ -58,6 +58,10 @@ impl<S: G> SealedSegment<S> {
     pub fn len(&self) -> u32 {
         self.indexing.len()
     }
+    pub fn size(&self) -> u64 {
```
usamoi, line 61: It will ignores the size of HNSW graph, quantization and IVF files.

src/index/views.rs
```diff
@@ -9,6 +9,7 @@ CREATE TYPE VectorIndexStat AS (
     idx_sealed BIGINT[],
     idx_growing BIGINT[],
     idx_write BIGINT,
+    idx_size BIGINT,
```
usamoi, line 12: We should document it.

crates/service/src/index/mod.rs
```diff
@@ -69,6 +69,7 @@ pub struct IndexStat {
     pub growing: Vec<u32>,
     pub write: u32,
     pub options: IndexOptions,
+    pub size: u64,
```
usamoi, line 72: It'll be better if we return info of all segments (id, type, length, size) directly.

src/index/views.rs
kemingy: fixed

crates/service/src/index/segments/sealed.rs
```diff
@@ -58,6 +58,22 @@ impl<S: G> SealedSegment<S> {
...
+            id: self.uuid,
+            typ: "sealed".to_string(),
+            length: self.len() as usize,
+            size: 0,
+        };
+        let path = self._tracker.path.join("indexing");
+        match dir_size(&path) {
+            Ok(size) => info.size = size as u64,
+            Err(e) => {
+                log::error!("Failed to get size of {:?}: {}", path, e);
```
usamoi, line 72: Panic might be better?

crates/service/src/utils/dir_ops.rs
```diff
@@ -1,7 +1,25 @@
...
 }
+
+pub fn dir_size(dir: &Path) -> io::Result<usize> {
+    let mut size = 0;
+    if dir.is_dir() {
+        for entry in read_dir(dir)? {
+            let entry = entry?;
+            let path = entry.path();
+            let name = path.file_name().unwrap().to_string_lossy();
+            if path.is_dir() && !name.starts_with('.') {
```
usamoi, line 17: Why `!name.starts_with('.')`?

src/index/views.rs
```diff
@@ -9,6 +9,7 @@ CREATE TYPE VectorIndexStat AS (
     idx_sealed BIGINT[],
     idx_growing BIGINT[],
     idx_write BIGINT,
+    idx_size TEXT,
```
usamoi, line 12: We can keep it as an integer. We just pass the `SegmentInfo` structure for implementation.

kemingy: What's the benefit?

crates/service/src/utils/dir_ops.rs
kemingy: Ignore the dir starts with `.`.

crates/service/src/index/segments/sealed.rs
kemingy: Collecting the OS file info may fail due to numerous reasons. Will the service panic be caught by pgrx and reflected on the postgres part?

usamoi: For the background worker process, if spawned threads panic, it just prints the log and exit the current session (so the transaction aborts). So it's OK to panic.

crates/service/src/utils/dir_ops.rs
usamoi: But isn't it not ignored in fact? It goes into `else` statement.

src/index/views.rs
usamoi: If `SegmentInfo` is exported in the future, it will be possible for users to get detailed information by reading JSON.

kemingy: Do you have any documents about querying int column as JSON?

crates/service/src/utils/dir_ops.rs
kemingy: Fixed.

src/index/views.rs
usamoi: I mean `stat` RPC might be exported as a Postgresql function returning JSON. So it will be possible to take advantages of Postgresql JSON functions.

kemingy: Got it. Fixed.

crates/service/src/index/segments/sealed.rs
usamoi: Anyway it should not return an incorrect size.

kemingy: Already changed to panic.

usamoi (approved):

PR #223, opened 2024-01-02
chore: friendly OOM message

### LINKED ISSUE #208: chore: Friendly OOM error
Currently backend daemon may panic when oom

usamoi:
I'm not sure if it really works. On my machine, OOM Killer kills Postgres directly when OOM occurs so there is no chance to print a nice message, but I think users can view kernel logs on this condition.

If OOM is detected by our code, only the OOM thread (probably an optimizing thread) will be killed. Then optimizing stops. Users will view our logs to know OOM occurs.

If the background process is killed but Postgres is not killed, the background process just restarts after 1 second. OOM probably occurs again (since it's raisen from an optimizing).

Will it close #208 ?

### DIFF
--- a/src/bgworker/mod.rs
+++ b/src/bgworker/mod.rs
@@ -4,12 +4,14 @@ pub mod upgrade;
 pub unsafe fn init() {
     use pgrx::bgworkers::BackgroundWorkerBuilder;
     use pgrx::bgworkers::BgWorkerStartTime;
+    use std::time::Duration;
     BackgroundWorkerBuilder::new("vectors")
         .set_function("vectors_main")
         .set_library("vectors")
         .set_argument(None)
         .enable_shmem_access(None)
         .set_start_time(BgWorkerStartTime::PostmasterStart)
+        .set_restart_time(Some(Duration::from_secs(1)))
         .load();
 }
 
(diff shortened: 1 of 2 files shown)


cutecutecat:
If the OOM is caused by `insert` instead of `optimize`, which thread would be killed, and would be restart?

usamoi:
@cutecutecat The session thread (the thread who communicate with a session process) panics and users can just restart the transaction (in theory).

src/bgworker/mod.rs
```diff
@@ -44,6 +53,9 @@ pub fn main() {
         }
         log::error!("Panickied. Info: {:?}. Backtrace: {}.", info, backtrace);
     }));
+    std::alloc::set_alloc_error_hook(|layout| {
```
cutecutecat, line 56: I have seen this API has been removed at https://github.com/rust-lang/rust/pull/109507 , and `OOM` is a regular panic from that.

Could https://doc.rust-lang.org/std/panic/fn.set_hook.html worked?

usamoi: If the pull request is merged, we will just remove set_alloc_hook and replace panic layload type to the std one. It's simple if we bump the compiler version someday.

cutecutecat: > If the pull request is merged, we will just remove set_alloc_hook and replace panic layload type to the std one. It's simple if we bump the compiler version someday.

Fine, Let's comment a TODO/Note at here to remind it.

VoVAllen (approved):

PR #224, opened 2024-01-03
feat: add optimizing_threads to gucs

### LINKED ISSUE #195: feat: Configure optimizing thread in runtime
It's currently in index build option, better to be able to change in runtime

cutecutecat:
Fix #195 
# Design

~~To use a `assign_hook` function, we export `unsafe DefineCustomIntVariable` from `pgrx`, instead of calling `GucRegistry::define_int_guc`.~~

[updated 2024/01/04]

I struggled with `assign_hook` and finally failed. To send RPC from hooked function would forbidden pg from starting. It seems import crate `let mut rpc = crate::ipc::client::borrow_mut();` caused that fault.

Now I use a simple Timer to sync `OPTIMIZING_THREADS` and call RPC.

We introduce 3 settings for this Callback feature:
- `RuntimeOptions`: A flattened Setting, Serializable for RPC
- `NotifiedSetting`: Arc\<Atomic\>, maintained by `Index`, could be restored from `IndexOptions`
- `SharedSetting`: Weak\<Atomic\>, maintained by `OptimizerIndexing`, downgraded from `NotifiedSetting`

The `assign_hook` would call RPC `setting` and set `NotifiedSetting` maintainered by `Index`, while optimizer would fetch `optimizing_threads` from a Weak reference `SharedSetting` from `NotifiedSetting.downgrade()`.

By this method, Arc could be fixed always in single process.

[updated 2024/01/05]
Now we have:
- GUCS `vectors.optimizing_threads_limit` for global limit for all indexes(Default 0 for no limit)
- Option `optimizing_threads` for each index

The threads of optimizing would be mininal of them, and if `vectors.optimizing_threads_limit==0`, there is no global limit.

### DIFF
--- a/crates/service/src/index/mod.rs
+++ b/crates/service/src/index/mod.rs
@@ -2,6 +2,7 @@ pub mod delete;
 pub mod indexing;
 pub mod optimizing;
 pub mod segments;
+pub mod setting;
 
 use self::delete::Delete;
 use self::indexing::IndexingOptions;
@@ -10,6 +11,8 @@ use self::segments::growing::GrowingSegment;
 use self::segments::growing::GrowingSegmentInsertError;
 use self::segments::sealed::SealedSegment;
 use self::segments::SegmentsOptions;
+use self::setting::NotifiedSetting;
+use self::setting::RuntimeOptions;
 use crate::index::optimizing::indexing::OptimizerIndexing;
 use crate::index::optimizing::sealing::OptimizerSealing;
 use crate::prelude::*;
@@ -99,6 +102,7 @@ pub struct Index<S: G> {
     view: ArcSwap<IndexView<S>>,
     instant_index: AtomicCell<Instant>,
     instant_write: AtomicCell<Instant>,
+    notified: NotifiedSetting,
     _tracker: Arc<IndexTracker>,
 }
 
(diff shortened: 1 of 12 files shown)


usamoi:
could you rebase?

usamoi (commented):
PostgreSQL is multiprocessing, so it's needed to use mmap to create shared memory between session process and background worker process if you want to use atomic instead of a RPC to implement it.

cutecutecat:
> PostgreSQL is multiprocessing, so it's needed to use mmap to create shared memory between session process and background worker process if you want to use atomic instead of a RPC to implement it.

Thank you for your carefully check, I have rewrite cross-process comunication with RPC, and now Atomic only used for cross-threads.

usamoi:
> P.S. I am not very sure whether background worker process could fetch GUCS, if there is still some error, please pointed out

The background worker process could fetch GUC values defined in `postgresql.conf` since the background process is a fork of postmaster.

usamoi:
It seems that users lost the ability to set optimizing threads for every index?

cutecutecat:
> It seems that users lost the ability to set number of optimizing threads for every index? Is it good? I think what we need is a global limit for all optimizing threads in all indexes.

Got it, so it need to be a seperated config from existing optimizing_threads? And would limit threads number of all the process.

VoVAllen:
We can keep it global now for simplicity.

cutecutecat:
> It seems that users lost the ability to set number of optimizing threads for every index? Is it good? I think what we need is a global limit for all optimizing threads in all indexes.

This fix contains a `optimizing_threads_limit` as a global limit, while we still keep `options.optimizing_threads` for set each index thread. PTAL @usamoi

usamoi:
I think what we need is a GUC `global_optimizing_threads` to limit threads of all optimizing threads.

If we modify the optimizing parameters, it's better if we persist them. GUC whose level is `Postmaster` are persisented by postgresql.

Issue #227, opened 2024-01-03
feat: Compatibility with pgvector


VoVAllen:
Support pgvector syntax out of the box

kemingy:
## Differences

| Type | `pgvector` | `pgvecto.rs` |
|--------|--------|--------|
| `<->` | L2 | Squared L2 |
| `<=>` | 1-cos | -cos |
| USING  | ivfflat/hnsw | vectors |
| operator | vector_ip_ops/vector_cosine_ops | vector_dot_ops/vector_cos_ops | 
| WITH | directly | `IndexOptions` toml |

I think the cosine distance should be `1-cos` instead of `cos`. Other differences look good to me. At least the pgvecto.rs indexing configuration looks better than pgvector.

cutecutecat:
After #256 , there is still a bit difference at Gucs, including:
- `hnsw.ef_search` at pgvector vs `vectors.hnsw_ef_search` at pgvecto.rs
- `ivfflat.probes` at pgvector vs `vectors.ivf_nporbe` at pgvecto.rs

PR #231, opened 2024-01-05
fix: cosine distance

kemingy:
- refer to https://github.com/tensorchord/pgvecto.rs/issues/227


gaocegege (dismissed):

VoVAllen:
How about PQ and SQ distance?

kemingy:
> How about PQ and SQ distance?

Fixed in the new commit. PTAL

usamoi (commented):
I suggest add a postgres_distance function to the `G` trait and use it for displaying.

kemingy:
> Can we just use `1.0-cos` and `sqrt(squared l2)` for PostgreSQL operator?

I don't have the plan to replace the squared L2 distance.

gaocegege:
Please resolve the conflicts.

> Can we just use `1.0-cos` and `sqrt(squared l2)` for PostgreSQL operator?

It will add more overhead to the compute, but we do not see the benefits. Thus let's keep no sqrt for now.

kemingy:
> Please resolve the conflicts.

fixed

gaocegege:
PTAL @usamoi @VoVAllen

usamoi (commented):
Please fix:
`https://github.com/tensorchord/pgvecto.rs/blob/a7a9a592c1ef8db0c41ff106b62c85359c10a3b1/README.md?plain=1#L109`
`https://github.com/tensorchord/pgvecto.rs/blob/a7a9a592c1ef8db0c41ff106b62c85359c10a3b1/README.md?plain=1#L118`
`https://github.com/tensorchord/pgvecto.rs/blob/a7a9a592c1ef8db0c41ff106b62c85359c10a3b1/crates/service/src/instance/metadata.rs#L21`

VoVAllen:
I think we can also add sqrt in L2 to make in consistent at all. The impact on performance should be negligible.

VoVAllen:
> Can we just use `1.0-cos` and `sqrt(squared l2)` for PostgreSQL operator?

I think this is the better way. We can add sqrt to PostgresSQL operator. If user didn't use L2 instance in the query, no extra cost will be introduced? And we can make it consistent with pgvector at all

usamoi:
> @usamoi Will the cost of sqrt happen when user doesn't need distance in the query?

No, there will be no extra cost. What I cares is consistency: we use `l2` for opeator name but it's `l2s`(`l2-squared`) in fact.

kemingy:
Cosine distance has to be fixed because it's defined as `1 - cosine_similarity`. For the L2 distance, if you choose to impl squared L2, it's also widely accepted. I don't think it's necessary to get the square root.

usamoi:
We didn't implement `Cosine distance`. What we implemented before is `negative cosine similarity` and we take `cos` for short. It's consistency with users who chooses to normalize the vector and use `negative dot product`.

usamoi (approved):

VoVAllen (approved):

Issue #215, opened 2023-12-29
chore: Unit test for more scenarios


VoVAllen:
Test:
- [x] Crash Recovery
- [x] Partition index (CREATE TABLE items (embedding vector(3), category_id int) PARTITION BY LIST(category_id);
- [x] Partial index (`CREATE INDEX ON items USING ivfflat (embedding vector_l2_ops) WITH (lists = 100) WHERE (category_id = 123);`)
- [x] Multiple index on single column
- [x] Reindex + Reindex Concurrently
- [x] Drop index when indexing

cutecutecat:
Discussion with @xieydd 

## Partition index
If two indexes are created at both child and parent table of `PARTITION`, which would be used in the query?

## Pgvector test format
pgvector uses `Tap Tests` to write tests not directly for `getting output`.
As for our `Concurrent` Test, it could not be easily handled by `sqllogictest`, as its target is output of single statement, not interaction.
> [Sqllogictest](https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki) is a testing framework to verify the correctness of an SQL database.

We need to discuss whether to introduce these framework.
ref: https://www.postgresql.org/docs/current/regress-tap.html

## Multiply index at column
If multiply indexes are created, which index would be used at query?

#

xieydd:
I think `reindex` and `reindex concurrently` is already supported. https://github.com/tensorchord/pgvecto.rs/blame/main/tests/sqllogictest/reindex.slt#L26

usamoi:
I have an idea that bgworker works independently so that we can just test bgworker without considering PostgreSQL stuff.

PR #242, opened 2024-01-08
chore: add crash test

cutecutecat:
Part of https://github.com/tensorchord/pgvecto.rs/issues/215

This is a test of Crash Recovery.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -87,7 +87,17 @@ jobs:
       - name: Sqllogictest
         run: |
           psql -f ./tests/init.sql
-          sqllogictest -u runner -d runner './tests/**/*.slt'
+          sqllogictest -u runner -d runner './tests/sqllogictest/*.slt'
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10' 
+      - name: Crash Recovery test
+        run: |
+          sudo pip install psutil==5.9.7
+          psql -f ./tests/init.sql
+          sqllogictest -u runner -d runner './tests/crash/create.slt'
+          sudo python ./tests/crash/kill.py
+          sqllogictest -u runner -d runner './tests/crash/restore.slt'
       - uses: actions/cache/save@v3
         with:
           path: |
--- a/tests/crash/kill.py
+++ b/tests/crash/kill.py
@@ -1,26 +1,27 @@
-from functools import reduce
 import logging
 import time
-from typing import Callable
 
 import psutil
 
 TIMEOUT = 30
-AND_FILTERS = ["postgres", "vectors"]
+FILTERS = ["postgres", "vectors"]
 logging.getLogger().setLevel(logging.INFO)
 
 
-def process_filter(p: psutil.Process):
+def process_filter(p: psutil.Process) -> bool:
     cmdline = "".join(p.cmdline())
-    filter_func: Callable[[bool, str], bool] = lambda ans, e: ans and (e in cmdline)
-    return reduce(filter_func, AND_FILTERS, True)
+    for case in FILTERS:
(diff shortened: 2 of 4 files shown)


gaocegege:
PTAL @usamoi

.github/workflows/check.yml
```diff
@@ -87,7 +87,17 @@ jobs:
...
+          sqllogictest -u runner -d runner './tests/sqllogictest/*.slt'
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10' 
+      - name: Crash Recovery test
+        run: |
+          sudo pip install psutil==5.9.7
+          psql -f ./tests/init.sql
+          sqllogictest -u runner -d runner './tests/crash/create.slt'
+          sudo python ./tests/crash/kill.py
```
kemingy, line 99: Why does python/pip require `sudo`?

tests/crash/kill.py
```diff
@@ -0,0 +1,54 @@
...
+
+def process_filter(p: psutil.Process):
+    cmdline = "".join(p.cmdline())
+    filter_func: Callable[[bool, str], bool] = lambda ans, e: ans and (e in cmdline)
+    return reduce(filter_func, AND_FILTERS, True)
+
+
+if __name__ == "__main__":
+    # Send kill signal to vectors process
+    timeout_start = time.time()
```
kemingy, line 21: 1. use monotonic time
2. `+ TIMEOUT` here

tests/crash/kill.py
```diff
@@ -0,0 +1,54 @@
...
+
+TIMEOUT = 30
+AND_FILTERS = ["postgres", "vectors"]
+logging.getLogger().setLevel(logging.INFO)
+
+
+def process_filter(p: psutil.Process):
+    cmdline = "".join(p.cmdline())
+    filter_func: Callable[[bool, str], bool] = lambda ans, e: ans and (e in cmdline)
+    return reduce(filter_func, AND_FILTERS, True)
```
kemingy, line 16: This is a bad usage of `reduce`. Could you try the normal code?

tests/crash/kill.py
```diff
@@ -0,0 +1,54 @@
...
+        time.sleep(1)
+
+    # Wait until process is not exist or recreated
+    timeout_start = time.time()
+    while True:
+        if time.time() > timeout_start + TIMEOUT:
+            raise TimeoutError(f"Background worker not killed in {TIMEOUT}s")
+        procs = [p for p in psutil.process_iter() if process_filter(p)]
+        pids = set(p.pid for p in procs)
+        if len(procs) == 0 or pids != last_pids:
```
kemingy, line 41: Do you mean `len(pids & last_pids) == 0`?

BTW, `last_pids` is kind of a temp var in the previous while loop.

.github/workflows/check.yml
cutecutecat: Kill process need a priviledge, and sudo python would use a seperated interpreter from local user. I understand sudo is soso, if there is a better way?

kemingy: It's necessary if the process is started with `sudo`.

tests/crash/kill.py
cutecutecat: Fixed.

tests/crash/kill.py
cutecutecat: Fixed.

tests/crash/kill.py
cutecutecat: Fixed, now extend last_pids's lifetime.

kemingy (approved):

usamoi (approved):

Issue #262, opened 2024-01-12
feat: Function / cast to convert vectors into vecf16


JoePassanante:
Hey folks,

My database currently has a 100k+ vectors, I need to reduce my memory usage before expanding further.   Ideally, I don't want to re compute these vectors from scratch nor write a script that moves that vectors into the vecf16 type. 

```SQL
ALTER TABLE "tableName"
ALTER COLUMN "embedding"
TYPE vecf16
USING embedding::vecf16
```

```
Error in query: ERROR: cannot cast type vector to vecf16
LINE 4: USING embedding::vecf16
```

Would it be possible to add an explicit cast, or function that allows this?

JoePassanante:
I think I can take this one. 

Going to introduce a new CAST 

```SQL
CREATE CAST (vector AS vecf16)
    WITH FUNCTION _vectors_cast_vecf32_to_vecf16(vector, integer, boolean) AS IMPLICIT;
```

and introduce a new fn in casts_f32.rs.

usamoi:
Tracking in #255.

Issue #255, opened 2024-01-11
feat: helper functions


usamoi:
pgvecto.rs are lack of helper functions, and we need to provide more helper functions

* function `vector_norm`: get length of a vector
* function `vector_dimensions`: get dimensions of a vector
* function `vector_normalize`: compute a normalized vector
* casts between `vector` and `vecf16`
* casts between `vecf16` and `half precision`

view https://docs.pgvecto.rs/reference/schema.html to see helper functions provided by pgvecto.rs so far

VoVAllen:
Test logical replication between different extension version

JoePassanante:
Can you elaborate on 
>  casts between vecf16 and half precision; PostgreSQL do not support half precision, we may make some decisions here

Is this an issue with postgres itself?  or that it hasn't been implememted in pgvecto?

usamoi:
> Is this an issue with postgres itself?

Yes. PostgreSQL supports floating-type `real` (IEEE 754 binary 32) and floating type `double precision` (IEEE 754 binary 64), but there is not such a type named `half precision` (IEEE 754 binary 16). We probably cannot provide a cast between a `vecf16` and an array. Another choice is providing a cast between `vef16` and `real[]`, which is "make some decisions here" refered.

JoePassanante:
> > Is this an issue with postgres itself?
> 
> Yes. PostgreSQL supports floating-type `real` (IEEE 754 binary 32) and floating type `double precision` (IEEE 754 binary 64), but there is not such a type named `half precision` (IEEE 754 binary 16). We probably cannot provide a cast between a `vecf16` and an array. Another choice is providing a cast between `vef16` and `real[]`, which is "make some decisions here" refered.

Messed over in the discord for some help getting the environment setup.  But isn't this as simple as:

```rust
#[pgrx::pg_extern(immutable, parallel_safe, strict)]
fn _vectors_cast_vecf32_to_vecf16(
    vector: Vecf32Input<'_>,
    _typmod: i32,
    _explicit: bool,
) -> Vecf16Output {
    let data = vector
        .data()
        .iter()
        .map(|x| x.to_f32())
        .map(|x| f16::from_f32(x))
        .collect();

    Vecf16::new_in_postgres(data)
}

CREATE CAST (vector AS vecf16)
    WITH FUNCTION _vectors_cast_vecf32_to_vecf16(vector, integer, boolean) AS IMPLICIT;
```

usamoi:
[...]

PR #266, opened 2024-01-13
feat: Add vector to vecf16 CAST

JoePassanante:
This adds the ability to cast vectors to vecf16.  Partially addressing #255 and closing #262, which specifically requests only this functionality.

### DIFF
--- a/tests/sqllogictest/cast.slt
+++ b/tests/sqllogictest/cast.slt
@@ -105,3 +105,9 @@ query I
 SELECT '{"[1,2,3]"}'::vector[];
 ----
 {"[1, 2, 3]"}
+
+# vector to vecf16
+query I
+SELECT '[1,]'::vector::vecf16;
+----
+[1]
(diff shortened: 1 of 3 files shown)


tests/sqllogictest/cast.slt
```diff
@@ -104,4 +104,14 @@ SELECT unnest('{"[1,2,3]", "[4,5,6]"}'::vector[]);
...
\ No newline at end of file
+{"[1, 2, 3]"}
+
+# vector to vecf16
+query I
+SELECT '[1,]'::vector::vecf16;
+----
+[1]
+
+# vecf16 to vector should fail
```
usamoi, line 115: We may provide a cast from `vecf16` to `vector` for convenience. Could you remove it?

JoePassanante: I personally like codifying negative cases, but its up to you.  

The reason why, at least initially, to do this is because I think going from vecf16 to vector might be a bad idea because you already lost that precision.

JoePassanante: Actually, I'll remove this and make the cast explicit.

JoePassanante:
@usamoi Do I need to bump the version myself?  or will you guys take care of that in another release cycle?

usamoi:
@JoePassanante You do not need to do that. Thank you for your contribution!

usamoi (approved):

VoVAllen:
@all-contributors please add @JoePassanante for code contribution

PR #273, opened 2024-01-15
chore: set schema and search_path

### LINKED ISSUE #270: bug: install at schema would fail at create index
Now we could not install pgvector.rs at schema.

```sql
CREATE SCHEMA tschema;
CREATE EXTENSION vectors WITH SCHEMA tschema;

CREATE TABLE tschema.t (val tschema.vector(3));
INSERT INTO tschema.t (val) SELECT ARRAY[random(), random(), random()]::real[] FROM generate_series(1, 1000);

CREATE INDEX native_index ON tschema.t USING vectors (val tschema.vector_cos_ops)
WITH (options = "[indexing.hnsw]");
ERROR:  type "vector" does not exist
```

Related function:
- pgrx::wrappers::regtypein("vector")
- regoperatorin("<#>(vector,vector)")

usamoi:
* pgvecto.rs: https://github.com/tensorchord/pgvecto.rs/pull/273
* pgvecto.rs-docs: https://github.com/tensorchord/pgvecto.rs-docs/pull/19

closes #270

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ version.workspace = true
 edition.workspace = true
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
 
 [features]
 default = ["pg15"]
--- a/docker/pgvecto-rs.Dockerfile
+++ b/docker/pgvecto-rs.Dockerfile
@@ -7,4 +7,4 @@ FROM tensorchord/pgvecto-rs-binary:${TAG}-${TARGETARCH} as binary
 FROM postgres:$POSTGRES_VERSION
 COPY --from=binary /pgvecto-rs-binary-release.deb /tmp/vectors.deb
 RUN apt-get install -y /tmp/vectors.deb && rm -f /tmp/vectors.deb
-CMD ["postgres", "-c" ,"shared_preload_libraries = 'vectors.so'", "-c", "search_path = 'vectors, \"\$user\", public'"]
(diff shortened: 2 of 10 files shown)


cutecutecat:
Is the `search_path` global and would be shared by other extensions?

cutecutecat:
Apart from this, I see there is a way of `An extension might be relocatable during installation but not afterwards.`
ref: https://www.postgresql.org/docs/16/extend-extensions.html#EXTEND-EXTENSIONS-RELOCATION

Could we set location at script as:
```
SET LOCAL search_path TO @extschema@, pg_temp;
```
And use a pg function `current_schema()` to get it at `regtypein` and `regoperatorin`?
ref: https://www.postgresql.org/docs/16/functions-info.html#FUNCTIONS-INFO-SESSION

usamoi:
> Is the `search_path` global and would be shared by other extensions?

`search_path` can be local.

usamoi:
> use a pg function `current_schema()` to get it at regtypein and regoperatorin?

I think it's not how `current_schema()` works.

cutecutecat (approved):
This PR would lock pgvecto.rs operators installed at `vectors` schema.

## Affect

`CREATE EXTENSION vectors WITH SCHEMA xxx` is forbidden from now.

Instead, user need to alter `search_path=vectors` to enable global access of operators and `search_path=vectors, xxx` to access other schema.

cc @VoVAllen

VoVAllen (approved):

PR #275, opened 2024-01-15
feat: FreshDiskANN

AveryQi115:
## TODO:
- [x] In memory vamana based on current segments
- [x] Vamana index with filtering
- [ ] Vamana index with deletion
- [ ] Optimize filtering algorithm

## Discussions:
- Current vamana index implementation use full precision building and searching, as current segments implementation is different from diskANN storage method, how to gradually build full-precision index stored in segments and using quantization to do the construction and searching process?

- performance and precision


marcusinthesky:
Really excited to see the amazing progress here, have you seen meillisearch's Arroy rust DiskANN implementation: https://github.com/meilisearch/arroy ?

PR #279, opened 2024-01-17
feat: transactional CREATE/DROP

usamoi:


VoVAllen:
Can we add some tests for this?

usamoi:
> Can we add some tests for this?

I do not know how to test it.

cutecutecat:
> > Can we add some tests for this?
> 
> I do not know how to test it.

Maybe we can start from what does this issue solved? Before and after?

usamoi:
> > > Can we add some tests for this?
> > 
> > 
> > I do not know how to test it.
> 
> Maybe we can start from what does this issue solved? Before and after?

The remote index created by CREATE will be removed if the transaction is aborted.

cutecutecat:
In previous discussion, if a user enter `ctrl+c` before SQL `CREATE INDEX` returned. At backend, `pgvecto.rs` will generate a `leaked` index which is never used and unmaintained forever.

This PR would trigger a index cleaner every time user send `ctrl+c`.

To test it, we might need a python script to send `abort` and check whether the index file is cleared soon later, which could be done later.

crates/service/src/index/mod.rs
cutecutecat, line 1: As `flush` caller is removed, do we still need `IndexView.flush()` and `InstanceView.flush()`?

usamoi: It's just moved to `Commit`.

VoVAllen (approved):

PR #287, opened 2024-01-19
fix: prevent overflow at kmeans

cutecutecat:
Thanks to @xieydd , we have digged out this bug.

```rust
Panickied. Info: PanicInfo { payload: Any { .. }, 
message: Some(attempt to subtract with overflow), 
location: Location { file: "crates/service/src/algorithms/clustering/[elkan_k_means.rs](http://elkan_k_means.rs/)", line: 179, col: 47 }, 
can_unwind: true, force_no_backtrace: false }
```
If there is too few vectors with a small count `n` and larger `nlist as c`, it will be a `overflow` there.

### DIFF
--- a/crates/service/src/algorithms/clustering/elkan_k_means.rs
+++ b/crates/service/src/algorithms/clustering/elkan_k_means.rs
@@ -83,6 +83,36 @@ impl<S: G> ElkanKMeans<S> {
         }
     }
 
+    /// Quick approach if we have little data
+    fn quick_centroids(&mut self) -> bool {
+        let c = self.c;
+        let samples = &self.samples;
+        let rand = &mut self.rand;
+        let centroids = &mut self.centroids;
+        let n = samples.len();
+        let dims = samples.dims();
+        let sorted_index = samples.argsort();
+        for i in 0..n {
+            let index = sorted_index.get(i).unwrap();
+            let last = sorted_index.get(std::cmp::max(i - 1, 0)).unwrap();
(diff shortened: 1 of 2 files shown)


usamoi:
should we just return centroids using a simpler algorithm if `n <= c`?

cutecutecat:
> should we just return centroids using a simpler algorithm if `n <= c`?

Fine, I will try `QuickCenters` by https://github.com/pgvector/pgvector/blob/cc9e6a67783092ae2e8b1b1ade552e547bef8934/src/ivfkmeans.c#L117

usamoi:
`QuickCenters` is already in the code.

crates/service/src/algorithms/clustering/elkan_k_means.rs
```diff
@@ -83,6 +83,36 @@ impl<S: G> ElkanKMeans<S> {
...
+        let c = self.c;
+        let samples = &self.samples;
+        let rand = &mut self.rand;
+        let centroids = &mut self.centroids;
+        let n = samples.len();
+        let dims = samples.dims();
+        let sorted_index = samples.argsort();
+        for i in 0..n {
+            let index = sorted_index.get(i).unwrap();
+            let last = sorted_index.get(std::cmp::max(i - 1, 0)).unwrap();
```
usamoi, line 97: Will there be integer overflow?

usamoi (approved):

PR #291, opened 2024-01-20
fix: correct fp16 dot operation

silver-ymz:


silver-ymz:
PTAL @usamoi @VoVAllen

usamoi (approved):

PR #256, opened 2024-01-12
feat: add compat with pgvector

cutecutecat:
Fix #227 

# Feat
- Add new Gucs `SET vectors.pgvector_compat=on` to enable/disable statement rewrite
- Function `pgvector_compat_rewrite` would be used to edit these elements
  - accessMethod -> from `hnsw/ivfflat` to `vectors`
  - opclass -> from `vector_ip_ops/vector_cosine_ops` to `vector_dot_ops/vector_cos_ops`
  - options -> from `m: int/ef_construction: int/list: int` to `options: serialized toml`
- Test file `compact.slt` for pgvector-style SQL statements

# Examples
These statments would be acceptable from now:
```sql
CREATE INDEX hnsw_l2_index ON t USING hnsw (val vector_l2_ops);
CREATE INDEX hnsw_ip_index ON t USING hnsw (val vector_ip_ops) WITH (ef_construction = 80);
CREATE INDEX hnsw_cosine_index ON t USING hnsw (val vector_cosine_ops) WITH (m = 12, ef_construction = 80);
CREATE INDEX ivf_l2_index ON t USING ivfflat (val vector_l2_ops) WITH (lists = 100);
```

# Limitation
There is still a bit difference at Gucs, including:
- `hnsw.ef_search` at pgvector vs `vectors.hnsw_ef_search` at pgvecto.rs
- `ivfflat.probes` at pgvector vs `vectors.ivf_nporbe` at pgvecto.rs

# Document
* https://github.com/tensorchord/pgvecto.rs-docs/pull/14

PR #296, opened 2024-01-22
feat: support pgvector gucs compact

### LINKED ISSUE #295: feat: guc compatibility with pgvector
Add guc compatibility with pgvector

cutecutecat:
Fix #295 
Related to #256 and #227
From now, we support `vectors.ivf_nprobe` and `vectors.hnsw_ef_search` like `pgvector` through overwrite.
However, their default value is not the same as pgvector.

# Document
* https://github.com/tensorchord/pgvecto.rs-docs/pull/14

### DIFF
--- a/src/gucs/executing.rs
+++ b/src/gucs/executing.rs
@@ -19,7 +19,7 @@ pub unsafe fn init() {
     );
     GucRegistry::define_int_guc(
         "vectors.ivf_nprobe",
-        "`nprobe` argument of HNSW algorithm.",
+        "`nprobe` argument of IVF algorithm.",
         "https://docs.pgvecto.rs/usage/search.html",
         &IVF_NPROBE,
         1,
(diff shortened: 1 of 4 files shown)


src/gucs/executing.rs
```diff
@@ -18,7 +18,7 @@ pub unsafe fn init() {
         GucFlags::default(),
     );
     GucRegistry::define_int_guc(
-        "vectors.ivf_nporbe",
+        "vectors.ivf_nprobe",
         "`nprobe` argument of HNSW algorithm.",
```
usamoi, line 22: Could you fix it by the way?

usamoi (approved):

src/gucs/executing.rs
usamoi, line 22:
```suggestion
        "`nprobe` argument of IVF algorithm.",
```

Issue #252, opened 2024-01-10
feat: sparse vector index support


VoVAllen:
Support sparse vector search such as generated by SPLADE

PR #299, opened 2024-01-23
feat: support sparse vector

silver-ymz:
part of #252

## Uasge

```sql
CREATE TABLE t (val svector(6));

INSERT INTO t (val) SELECT ARRAY[0, random(), 0, 0, random(), random()]::real[]::vector::svector FROM generate_series(1, 1000);

INSERT INTO t (val) SELECT "[0, 1, 1, 0, 0, 0]";

CREATE INDEX ON t USING vectors (val svector_l2_ops)
WITH (options = "[indexing.hnsw]");

CREATE INDEX ON t USING vectors (val svector_dot_ops)
WITH (options = "[indexing.hnsw]");

CREATE INDEX ON t USING vectors (val svector_cos_ops)
WITH (options = "[indexing.ivf]");
```

### Storage

The storage part is in `crates/service/src/prelude/storage`.

`DenseMmap` is original `RawMmap`, sparse storage struct is following
```rust
pub struct SparseMmap {
    vectors: MmapArray<SparseF32Element>,
    offsets: MmapArray<u32>,
    payload: MmapArray<Payload>,
    dims: u16,
}
```

### Index

`Hnsw` doesn't make any modification. For `ivf`, it expands all vectors to dense vector as the subsamples of kmeans, and all kmeans algorithm is used on the dense vector. After getting centroids, it will calculate other vectors in sparse format.

`SQ` and `PQ` aren't implemented now.


silver-ymz:
PTAL @VoVAllen @usamoi

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,321 @@
...
+use std::ffi::CStr;
+use std::ffi::CString;
+use std::ops::Deref;
+use std::ops::DerefMut;
+use std::ops::Index;
+use std::ops::IndexMut;
+use std::ptr::NonNull;
+
+#[repr(C, align(8))]
+pub struct SVecf32 {
```
usamoi, line 23: We may need to track dimensions of a vector.

crates/service/src/algorithms/flat.rs
```diff
@@ -56,64 +45,80 @@ impl<S: G> Flat<S> {
...
+
+    pub fn dims(&self) -> u16 {
+        self.mmap.raw.dims()
+    }
+
+    pub fn len(&self) -> u32 {
+        self.mmap.raw.len()
+    }
+
+    pub fn content(&self, i: u32) -> &[S::Element] {
```
usamoi, line 63: Too. We can add a associcated type `vector<'a>` to `G`.

src/datatype/svecf32.rs
silver-ymz: Current implementation is to ignore dimensions in `SVef32` to support dynamic dimension like `vector`. When casting to `vector`, users must set the dimension explicitly. And when passing a sparse vector to the backend, it will check the last element.

https://github.com/silver-ymz/pgvecto.rs/blob/feat/sparse/crates/service/src/prelude/storage/sparse.rs#L54-L56

crates/service/src/algorithms/flat.rs
silver-ymz: There is a function to expand a sparse vector.

https://github.com/silver-ymz/pgvecto.rs/blob/feat/sparse/crates/service/src/prelude/storage/mod.rs#L32

src/datatype/svecf32.rs
usamoi: I think it's not a good way. Let's record dimensions of a vector in this structure. We can still implement "dynamic dimension" as long as we do not check typmod. It's consistent with semantics of `vector`.

crates/service/src/algorithms/flat.rs
usamoi: I prefer such a structure:
```rust
struct Vecf32<'a> {
    dimensions: &'a [F32],
}
struct SVecf32<'a> {
    dims: u16,
    elements: &'a [(u16, F32)],
}
```
It will make programming easier.

src/datatype/svecf32.rs
silver-ymz: added in new commit

crates/service/src/algorithms/flat.rs
silver-ymz: added in new commit. But I think it seems to introduce more intermediate types, making programming more complex.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,326 @@
...
+    varlena: u32,
+    len: u16,
+    kind: u8,
+    reserved: u8,
+    dims: u16,
```
usamoi, line 28: There is a pair of padding bytes and we need to fill them with zeros. It's required by PostgreSQL.

silver-ymz: completed in new commit

crates/service/src/prelude/global/mod.rs
```diff
@@ -2,22 +2,39 @@ mod f16;
...
 pub use f32_l2::F32L2;
+pub use sparse_f32_cos::SparseF32Cos;
+pub use sparse_f32_dot::SparseF32Dot;
+pub use sparse_f32_l2::SparseF32L2;
 
 use crate::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Display};
 
 pub trait G: Copy + Debug + 'static {
```
usamoi, line 28: add `type Vector<'a>: ...` here

crates/service/src/prelude/scalar/sparse_f32.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn vector<'a, 'b>(&'a self) -> &'b [Self::Element]
+    where
+        'c: 'b,
+    {
+        self.elements
+    }
+}
+
+pub fn expand_sparse(sparse: &[SparseF32Element]) -> impl Iterator<Item = F32> + '_ {
+    let mut data = RefPeekable::new(sparse.iter());
```
usamoi, line 60:
```suggestion
    let mut data = sparse.iter().peekable();
```
We removed `RefPeekable` in another pending PR.

crates/service/src/prelude/storage/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn dims(&self) -> u16;
+    fn len(&self) -> u32;
+    fn content(&self, i: u32) -> &[Self::Element];
+    fn payload(&self, i: u32) -> Payload;
+}
+
+pub trait Storage {
+    type Element: Copy + bytemuck::Pod;
+    type Scalar: Copy;
+    type Vector: Vector<Element = Self::Element>;
```
usamoi, line 24: Move these assiocated types to trait `G`.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,490 @@
...
+        (size << 2) as u32
+    }
+    fn layout(len: usize) -> Layout {
+        u16::try_from(len).expect("Vector is too large.");
+        let layout_alpha = Layout::new::<SVecf32>();
+        let layout_beta = Layout::array::<SparseF32Element>(len).unwrap();
+        let layout = layout_alpha.extend(layout_beta).unwrap().0;
+        layout.pad_to_align()
+    }
+    pub fn new_in_postgres(slice: &[SparseF32Element], dims: u16) -> SVecf32Output {
```
usamoi, line 44: We must **not** use `&[SparseF32Element]` everywhere. We must know `dims`. Use the type behind  `G::VectorRef<'a>`.

crates/service/src/algorithms/quantization/product.rs
```diff
@@ -122,34 +122,41 @@ impl<S: G> Quan<S> for ProductQuantization<S> {
...
-        raw: &Raw<S>,
+        raw: &Raw<S::Storage>,
         normalizer: F,
     ) -> Self
     where
         F: Fn(u32, &mut [S::Scalar]),
     {
-        std::fs::create_dir(&path).unwrap();
+        debug_assert!(
+            options.vector.k != Kind::SparseF32,
```
usamoi, line 135: Use `G::DISTANCE`.

crates/service/src/algorithms/quantization/product.rs
```diff
@@ -122,34 +122,41 @@ impl<S: G> Quan<S> for ProductQuantization<S> {
...
         quantization_options: QuantizationOptions,
-        raw: &Raw<S>,
+        raw: &Raw<S::Storage>,
         normalizer: F,
     ) -> Self
     where
         F: Fn(u32, &mut [S::Scalar]),
     {
-        std::fs::create_dir(&path).unwrap();
+        debug_assert!(
```
usamoi, line 134:
```suggestion
        assert!(
```

src/prelude/error.rs
```diff
@@ -70,6 +70,10 @@ INFORMATION: left_dimensions = {left_dimensions}, right_dimensions = {right_dime
         left_dimensions: u16,
         right_dimensions: u16,
     },
+    #[error("{message}")]
+    Custom {
```
usamoi, line 74: Remove it.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,490 @@
...
+    while index < vector.dims() {
+        buffer.push_str(", 0");
+        index += 1;
+    }
+    buffer.push(']');
+    CString::new(buffer).unwrap()
+}
+
+#[pgrx::pg_extern(immutable, parallel_safe, strict)]
+fn svector_from_kv_string(dims: i32, input: &str) -> SVecf32Output {
```
usamoi, line 338: Do not expose public functions by `pg_extern`. Write schema manually.

crates/service/src/prelude/storage/sparse.rs
```diff
@@ -0,0 +1,85 @@
...
+            dims: self.dims,
+            elements: &self.vectors[s..e],
+        }
+    }
+
+    fn payload(&self, i: u32) -> Payload {
+        self.payload[i as usize]
+    }
+
+    fn full_vector(contents: Self::VectorRef<'_>) -> Cow<'_, [Self::Scalar]> {
```
usamoi, line 41: Remove it.

crates/service/src/algorithms/quantization/scalar.rs
```diff
@@ -40,18 +40,23 @@ impl<S: G> ScalarQuantization<S> {
...
     fn create(
-        path: PathBuf,
+        path: &Path,
         options: IndexOptions,
         _: QuantizationOptions,
-        raw: &Arc<Raw<S>>,
+        raw: &Arc<Raw<S::Storage>>,
     ) -> Self {
-        std::fs::create_dir(&path).unwrap();
+        debug_assert!(
```
usamoi, line 48: Where is feature gate? I didn't see it in a `impl Validate`.

crates/service/src/prelude/global/mod.rs
silver-ymz: completed in new commit

crates/service/src/prelude/scalar/sparse_f32.rs
silver-ymz: completed in new commit

crates/service/src/prelude/storage/mod.rs
silver-ymz: completed in new commit

src/datatype/svecf32.rs
silver-ymz: completed in new commit

crates/service/src/algorithms/quantization/product.rs
silver-ymz: completed in new commit

crates/service/src/algorithms/quantization/product.rs
silver-ymz: completed in new commit

src/prelude/error.rs
silver-ymz: completed in new commit

crates/service/src/algorithms/quantization/scalar.rs
silver-ymz: Previously, it was checked in `crates/service/src/instance/mod.rs` manually. Now, it tells `validator_derive` to use `validate_index_options` function.

usamoi:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
pub struct SparseF32Element {
    pub index: u32,
    pub value: F32,
}
```

It'll be better if we use two arrays (array of `index: u16` and `value: F32`). We'll save 25% memory on vectors.

VoVAllen:
Probably need some experiments to see the best calculation method/storage layout. In pyanns, it stored the data along with the indices like [1, 0.1, 2, 0.1]. Therefore the read is totally sequential. Seperate them needs a seperate random access.

VoVAllen:
Also there should be chances to use simd to optimize the computation. if statement inside calculation should be much slower

usamoi:
> Seperate them needs a seperate random access

We will read both two arrays at the same time so it's still sequential.

> there should be chances to use simd to optimize the computation

This layout should be nice for SIMD.

VoVAllen:
I'm actually thinking of bitset like encoding for indices such as roaringbitmap. Then you can get the intersection of the set easily, and then select out the value to compute. Not sure how much improvement we can have here. Is it worth trying?

usamoi:
> I'm actually thinking of bitset like encoding for indices such as roaringbitmap. Then you can get the intersection of the set easily, and then select out the value to compute. Not sure how much improvement we can have here. Is it worth trying?

Roaring bitmap is designed for a large data. How could we take advantage of it?

VoVAllen:
I just feel bitset-like structure is easier to use SIMD. Find this might be useful as a referenece https://github.com/UNSW-database/simd_set_operations

src/datatype/svecf32.rs
silver-ymz: completed in new commit

silver-ymz:
It's ready. @usamoi

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,698 @@
...
+use std::ptr::NonNull;
+
+#[repr(C, align(8))]
+pub struct SVecf32 {
+    varlena: u32,
+    len: u16,
+    kind: u8,
+    reserved: u8,
+    dims: u16,
+    phantom: [u8; 0],
```
usamoi, line 27: Add padding bits to force alignment to be 8.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,698 @@
...
+    }
+    buffer.push(']');
+    CString::new(buffer).unwrap()
+}
+
+#[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16"))]
+#[pgrx::pg_extern(sql = "\
+CREATE FUNCTION _vectors_svecf32_subscript(internal) RETURNS internal
+IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '@FUNCTION_NAME@';")]
+fn _vectors_svecf32_subscript(_fcinfo: pgrx::pg_sys::FunctionCallInfo) -> Datum {
```
usamoi, line 330: Copy the test file.

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
silver-ymz: Previously, there is `tests/sqllogictest/pg14/svector_subscript.slt`.

VoVAllen:
We can possibly convert the computation to branchless like https://lemire.me/blog/2021/07/14/faster-sorted-array-unions-by-reducing-branches/

src/sql/finalize.sql
```diff
@@ -241,11 +349,24 @@ CREATE OPERATOR <=> (
...
+	RIGHTARG = svector,
+	COMMUTATOR = <=>
+);
+
 -- List of functions
 
 CREATE FUNCTION pgvectors_upgrade() RETURNS void
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_pgvectors_upgrade_wrapper';
 
+CREATE FUNCTION svector_from_split_array("dims" INT, "index" INT[], "value" real[]) RETURNS svector
```
VoVAllen, line 364:
```suggestion
CREATE FUNCTION svector_from_array("dims" INT, "index" INT[], "value" real[]) RETURNS svector
```

VoVAllen: And please check and force the index to be sorted (monotonically increasing), raise exception otherwise.

src/sql/finalize.sql
```diff
@@ -241,11 +349,24 @@ CREATE OPERATOR <=> (
...
+IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_svector_from_split_array_wrapper';
+
+CREATE FUNCTION svector_from_kv_string("dims" INT, "input" TEXT) RETURNS svector
```
VoVAllen, line 367:
```suggestion
CREATE FUNCTION svector_from_kv("dims" INT, "input" TEXT) RETURNS svector
```

VoVAllen: Remember to raise error if duplicate key found

src/sql/finalize.sql
silver-ymz: The current implementation will sort arrays and check whether the lengths of two arrays are equal, whether either array contains null, and whether there are duplicate keys.

https://github.com/tensorchord/pgvecto.rs/blob/65b83f0c2f3f20d252ce1af640d171266813e50e/src/datatype/svecf32.rs#L709-L770

VoVAllen:
pg supports function overloading, I think we can use the same function name `to_svector` to follow pg conventions

https://www.postgresql.org/docs/current/xfunc-overload.html

src/sql/finalize.sql
silver-ymz: Already checked

https://github.com/tensorchord/pgvecto.rs/blob/65b83f0c2f3f20d252ce1af640d171266813e50e/src/datatype/svecf32.rs#L677-L687

silver-ymz:
> We can possibly convert the computation to branchless like https://lemire.me/blog/2021/07/14/faster-sorted-array-unions-by-reducing-branches/

It seems that `if cond {} else {}` won't use conditional-move instructions. But we can use something like `xy += ((lhs_index == rhs_index) as u32 as f32) * lhs_value * rhs_value;`

compiler results for the first case:
https://godbolt.org/z/1zhGG8o79

compiler results for the second case:
https://godbolt.org/z/exo3j79eh

VoVAllen:
Do we need to store the vector norm for L2/cos distance in sparse vector? @usamoi

VoVAllen:
We can handle the branchless/SIMD optimization in other PR since it doesn't affect internal layout

src/sql/finalize.sql
VoVAllen: LGTM

VoVAllen:
> I just feel bitset-like structure is easier to use SIMD. Find this might be useful as a referenece https://github.com/UNSW-database/simd_set_operations , with his [thesis](https://github.com/UNSW-database/simd_set_operations/blob/main/thesis.pdf) including thorough analysis

After reading the thesis, I think our final target should be https://github.com/UNSW-database/simd_set_operations/blob/main/setops/src/intersect/broadcast.rs#L489-L561. On AVX512 machine, it's pretty competitive with the array representation.

usamoi:
Do we have the reference of format for `svector_from_kv_string`?

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+        if dims == 0 || len == 0 {
+            pgrx::error!("data corruption is detected");
+        }
+        let indexes_bytes = std::mem::size_of::<u16>() * len as usize;
+        let indexes_ptr = pgrx::pg_sys::pq_getmsgbytes(buf, indexes_bytes as _);
+        let values_bytes = std::mem::size_of::<F32>() * len as usize;
+        let values_ptr = pgrx::pg_sys::pq_getmsgbytes(buf, values_bytes as _);
+        let mut output = SVecf32::new_zeroed_in_postgres(len as usize);
+        output.dims = dims;
+        std::ptr::copy(indexes_ptr, output.indexes_mut() as _, indexes_bytes);
```
usamoi, line 588: Let's detect possible data corruption for indexes.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+    }
+    fn values(&self) -> *const F32 {
+        let len = self.len as usize;
+        unsafe {
+            let ptr = self.phantom.as_ptr().cast::<u16>().add(len);
+            let offset = ptr.align_offset(8);
+            ptr.add(offset).cast()
+        }
+    }
+    fn indexes_mut(&mut self) -> *mut u16 {
```
usamoi, line 101: Remove it.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+        self.phantom.as_mut_ptr().cast()
+    }
+    fn values_mut(&mut self) -> *mut F32 {
```
usamoi, line 104: Remove it.

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+    pub fn dims(&self) -> u16 {
+        self.dims
+    }
+    pub fn len(&self) -> usize {
+        self.len as usize
+    }
+    fn indexes(&self) -> *const u16 {
+        self.phantom.as_ptr().cast()
+    }
+    fn values(&self) -> *const F32 {
```
usamoi, line 93: Why not return a reference?

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+            SVecf32Output(NonNull::new(ptr).unwrap())
+        }
+    }
+    pub fn dims(&self) -> u16 {
+        self.dims
+    }
+    pub fn len(&self) -> usize {
+        self.len as usize
+    }
+    fn indexes(&self) -> *const u16 {
```
usamoi, line 90: Why not return a reference?

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+            }
+        }
+    }
+    if state != MatchedRight {
+        SessionError::BadLiteral {
+            hint: "Bad sequence.".to_string(),
+        }
+        .friendly();
+    }
+    SVecf32::new_in_postgres(SparseF32Ref {
```
usamoi, line 324: We need to check if the dimensions are in `[1, 65535]`.

src/prelude/error.rs
```diff
@@ -70,6 +70,14 @@ INFORMATION: left_dimensions = {left_dimensions}, right_dimensions = {right_dime
         left_dimensions: u16,
         right_dimensions: u16,
     },
+    #[error("\
```
usamoi, line 73: Use `BadLiteral`.

crates/service/src/index/indexing/mod.rs
```diff
@@ -130,11 +161,11 @@ impl<S: G> DynamicIndexing<S> {
         }
     }
 
-    pub fn vector(&self, i: u32) -> &[S::Scalar] {
+    pub fn content(&self, i: u32) -> S::VectorRef<'_> {
```
usamoi, line 164: Let's keep the name of `vector` and `open`. Afterall we already use the two names everywhere.

VoVAllen:
> Do we have the reference of format for `svector_from_kv_string`?

I don't  think we need from kv string capability now. User can easily transform kv to the array in any language. The kv format I think is rare

VoVAllen:
And I think we don't need to worry about the sparse computation now. We can always improve it later

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
silver-ymz: `_vectors_svecf32_recv` needs it. I updated the returned value from a pointer to a reference.

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
silver-ymz: done in new commit

src/prelude/error.rs
silver-ymz: done in new commit

src/datatype/svecf32.rs
```diff
@@ -0,0 +1,770 @@
...
+use std::ffi::CStr;
+use std::ffi::CString;
+use std::ops::Deref;
+use std::ops::DerefMut;
+use std::ptr::NonNull;
+
+#[repr(C, align(8))]
+pub struct SVecf32 {
+    varlena: u32,
+    len: u16,
```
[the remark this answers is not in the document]
silver-ymz: done in new commit

crates/service/src/index/indexing/mod.rs
```diff
@@ -114,11 +119,37 @@ impl<S: G> DynamicIndexing<S> {
...
+        filter: impl Filter + 'a,
+    ) -> (Vec<Element>, Box<(dyn Iterator<Item = Element> + 'a)>) {
+        match self {
+            DynamicIndexing::Flat(x) => x.vbase(vector, opts, filter),
+            DynamicIndexing::Ivf(x) => x.vbase(vector, opts, filter),
+            DynamicIndexing::Hnsw(x) => x.vbase(vector, opts, filter),
+        }
+    }
+
+    pub fn dims(&self) -> u16 {
```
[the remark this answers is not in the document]
silver-ymz: removed in new commit

crates/service/src/index/indexing/mod.rs
silver-ymz: done in new commit

src/sql/finalize.sql
```diff
@@ -241,11 +349,21 @@ CREATE OPERATOR <=> (
...
+	RIGHTARG = svector,
+	COMMUTATOR = <=>
+);
+
 -- List of functions
 
 CREATE FUNCTION pgvectors_upgrade() RETURNS void
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_pgvectors_upgrade_wrapper';
 
+CREATE FUNCTION to_svector("dims" INT, "index" INT[], "value" real[]) RETURNS svector
```
VoVAllen, line 364:
```suggestion
CREATE FUNCTION to_svector("dims" INT, "indices" INT[], "values" real[]) RETURNS svector
```

crates/service/src/prelude/scalar/sparse_f32.rs
```diff
@@ -0,0 +1,87 @@
+use crate::prelude::*;
+use serde::{Deserialize, Serialize};
+use std::fmt::Display;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SparseF32Element {
```
usamoi, line 6: Will remove it.

usamoi (approved):

usamoi:
Please fix the implementation of comparsion operators.

Issue #115, opened 2023-11-03
Diesel Support


kj3moraes:
Will there be support added for [Diesel](https://github.com/diesel-rs/diesel) ? It is easier to use Diesel to generate SQL queries.

cutecutecat:
Diesel only supports binary format https://github.com/diesel-rs/diesel/issues/2087 now and it seems unchanged yet.
But Pgvecto.rs not support binary result format now.
Might related: https://github.com/tensorchord/pgvecto.rs/issues/100
For now, You would meet `ERROR:  no binary output function available for type vector`.

Here is a partial finished example for `Diesel`, `insert` is fine while `select` is still ill.
```golang
#[macro_use]
extern crate diesel;

use std::{env, vec};

use chunks::emb;
use diesel::expression::{AsExpression, Expression, TypedExpressionType};
use diesel::{insert_into, Connection, PgConnection, QueryDsl, RunQueryDsl};
use diesel::{pg::Pg, sql_types::*};

diesel::infix_operator!(Emb, " <-> ", backend: Pg);

pub fn dis<T, U, ST>(left: T, right: U) -> Emb<T, U::Expression>
where
    T: Expression<SqlType = ST>,
    U: AsExpression<ST>,
    ST: SqlType + TypedExpressionType,
{
    Emb::new(left, right.as_expression())
}

table! {
    chunks {
        id -> Integer,
        emb -> Array<Float>,
    }
}

#[derive(Insertable, Queryable, Identifiable, Debug, PartialEq, Selectable)]
#[diesel(table_name = chunks)]
#[diesel(check_for_backend(Pg))]
struct VectorTable {
    id: i32,
    emb: Vec<f32>,
}

fn main() {
    // let database_url = env::var("DATABASE_URL").expect("DATABASE_URL must be set");
    let database_url = "postgresql://postgres:mysecretpassword@localhost:5432".to_string();
    let mut connection = PgConnection::establish(&database_url)
        .unwrap_or_else(|_| panic!("Error connecting to {}", database_url));

    let values: Vec<VectorTable> = vec![
        VectorTable {
            id: 0,
            emb: vec![0.1, 0.2, 0.3],
        },
        VectorTable {
            id: 1,
            emb: vec![0.1, 0.2, 0.3],
        },
    ];
    let _ = insert_into(chunks::table)
        .values(&values)
        .execute(&mut connection)
        .expect("insert failed");
    let filter = chunks::table
        .select(emb)
[...]

PR #314, opened 2024-01-29
feat: support binary representation

usamoi:
DO NOT MERGE IT BEFORE `0.2.0` IS RELEASED


cutecutecat:
I see pgvector use `copy binary` to test it.

https://github.com/pgvector/pgvector/blob/c22740962cda6a70ebb7a4f30d7970ff65eaef6e/test/sql/copy.sql

VoVAllen:
How to utilize this feature in Python SDK?

usamoi:
> How to utilize this feature in Python SDK?

I know nothing about Python SDK.

VoVAllen:
https://www.postgresql.org/docs/current/xtypes.html

Seems it only works for copy command. Probably need to convert the vectors into binary representation and then send it through copy command in Python

VoVAllen:
Is it exactly the same as our in-memory representation for each type?

usamoi:
> Is it exactly the same as our in-memory representation for each type?

No. We use a simpler representation. The in-memory representation contains PostgreSQL internals.

usamoi:
> https://www.postgresql.org/docs/current/xtypes.html
> 
> Seems it only works for copy command. Probably need to convert the vectors into binary representation and then send it through copy command in Python

There should be a way to use binary repr directly in a connection because Diesel only accepts binary repr (#115).

VoVAllen:
Can you try to write some python code to see how to convert a numpy array to it?

Possible example using copy:
```python
import psycopg2
import io

# Your database connection details
conn = psycopg2.connect("dbname=your_db user=your_user password=your_password")
cur = conn.cursor()

# Your list of tuples
data_list = [(1, 'Alice'), (2, 'Bob'), (3, 'Charlie')]

# Create a StringIO object and write your list of tuples to it as a CSV
output = io.StringIO()
for row in data_list:
    output.write('|'.join(map(str, row)) + '\n')  # Using | as a delimiter, for example
output.seek(0)  # Be sure to reset the position to the start of the StringIO object

# Prepare your COPY command
PGVECTOR_INDEX = 'your_table_name'
col_name_tuple = ['column1', 'column2']  # Column names in your table
copy_command = f"COPY {PGVECTOR_INDEX} ({', '.join(col_name_tuple)}) FROM STDIN WITH CSV DELIMITER '|'"

# Use copy_expert to load the data
cur.copy_expert(copy_command, output)

# Commit the transaction and close the connection
conn.commit()
cur.close()
conn.close()

```

usamoi:
> Can you try to write some python code to see how to convert a numpy array to it?
> 
> Possible example using copy (by ChatGPT):
> 
> ```python
> import psycopg2
> import io
> from psycopg2.extras import execute_values
> 
> # Your database connection details
> conn = psycopg2.connect("dbname=your_db user=your_user password=your_password")
> cur = conn.cursor()
> 
> # Your list of tuples
> data_list = [(1, 'Alice'), (2, 'Bob'), (3, 'Charlie')]
> 
> # Use a BytesIO buffer since we are dealing with binary data
> buffer = io.BytesIO()
> 
> # Write the magic word for Postgres binary format
> buffer.write(b'PGCOPY\n\xff\r\n\0')
> buffer.write(b'\0\0\0\0\0\0\0\0')  # Flags field
> buffer.write(b'\0\0\0\0')  # Header extension area length
> 
> # Write the data to the buffer
> for row in data_list:
>     buffer.write(b'\0\0\0\2')  # Indicates that there are two columns (adjust as needed)
>     
>     for item in row:
>         # Convert the item to bytes and write it to the buffer
>         if isinstance(item, int):
>             # Integer
>             data = item.to_bytes(4, byteorder='big', signed=True)
>             buffer.write(b'\0\0\0\4' + data)
>         elif isinstance(item, str):
>             # String
>             data = item.encode('utf-8')
>             buffer.write(len(data).to_bytes(4, byteorder='big', signed=True) + data)
>     
> buffer.write(b'\xff\xff')  # File trailer
> 
> buffer.seek(0)  # Move the buffer cursor to the start
> 
> # Prepare your COPY command
> PGVECTOR_INDEX = 'your_table_name'
> col_name_tuple = ['column1', 'column2']  # Column names in your table
> copy_command = f"COPY {PGVECTOR_INDEX} ({', '.join(col_name_tuple)}) FROM STDIN WITH BINARY"
> 
> # Use copy_expert to load the data
> cur.copy_expert(copy_command, buffer)
> 
> # Commit the transaction and close the connection
> conn.commit()
> cur.close()
> conn.close()
> ```
> 
> > In this script:
> > The magic word 'PGCOPY\n\xff\r\n\0' is necessary to identify the format as binary copy format.
> > The next 8 bytes are reserved for flags (currently must be zero).
> > The next 4 bytes are reserved for the length of the header extension (currently must be zero).
> > For each field (column value), the length of the field (in bytes) is specified followed by the field data itself.
> > After all fields of a row are copied, the next row starts immediately with the length of its first field.
> > At the end of the data, the special sequence '\xff\xff' indicates the end of the copy data.
> 
> Other reference: https://nickb.dev/blog/disecting-the-postgres-bulk-insert-and-binary-format/

I think psycopg3 supports it natively.

usamoi:
PTAL @VoVAllen

VoVAllen:
Have you tested with Python?

usamoi:
> Have you tested with Python?

No. Python packages does not support binary format well. But I have tested it with `libpq`.

Issue #318, opened 2024-01-30
bug: error when reindexing after upgrading to 0.2


mertalev:
When calling `REINDEX INDEX` on an HNSW index after upgrading and restarting, it throws the error `internal error: entered unreachable code`. Restarting Postgres again doesn't seem to have an effect. Brute-force search still works.

Steps leading up to this:
```sql
START TRANSACTION;

ALTER DATABASE immich SET search_path TO "$user", public, vectors;
CREATE SCHEMA IF NOT EXISTS vectors; -- this step is required for the ALTER EXTENSION line to work
UPDATE pg_catalog.pg_extension SET extversion = '0.1.11' WHERE extname = 'vectors';
UPDATE pg_catalog.pg_extension SET extrelocatable = true WHERE extname = 'vectors';
ALTER EXTENSION vectors SET SCHEMA vectors;
UPDATE pg_catalog.pg_extension SET extrelocatable = false WHERE extname = 'vectors';
ALTER EXTENSION vectors UPDATE TO '0.2.0-alpha.1';
SELECT pgvectors_upgrade();

COMMIT;

-- throws an error "The extension is upgraded so all index files are outdated." - not sure if intended behavior
SELECT idx_status FROM pg_vector_index_stat WHERE indexname = 'clip_index'; 
```
After restarting:
```sql
ALTER DATABASE immich SET search_path TO "$user", public, vectors;

-- throws a different error "The index is not existing in the background worker."
-- not sure if intended behavior, but we handle this error and continue
SELECT idx_status FROM pg_vector_index_stat WHERE indexname = 'clip_index'; 

REINDEX INDEX clip_index; -- throws 'internal error: entered unreachable code'
```
I also see instances of this in the logs:
```
ERROR:  pgvecto.rs: The index is not existing in the background worker.
ADVICE: Drop or rebuild the index.
CONTEXT:  while vacuuming index "face_index" of relation "public.asset_faces"`
```
This log doesn't show up for `clip_index`, only for `face_index`. The app doesn't call `REINDEX INDEX face_index` since doing this for `clip_index` throws an error first, so my hunch was that it was related to this. But these logs continue to appear after manually calling `REINDEX INDEX face_index`.

Calling `SELECT idx_status FROM vectors.pg_vector_index_stat` throws ```ERROR: called `Result::unwrap()` on an `Err` value: NoSuchType("vector_index_stat")```

[...]

PR #319, opened 2024-01-30
fix: add schema for referring vector_index_stat

usamoi:
in #318


VoVAllen (dismissed):

VoVAllen:
Why is this not covered by the test?

usamoi:
> Why is this not covered by the test?

It's covered now.

PR #328, opened 2024-01-31
refactor: faster IVF & PQ

whateveraname:
### Work done
- add multi-threading support for IVF index building and K-Means 
- fix a bug where IVFPQ are not trained with IVF residuals
- store codes in the layout that codes in the same cluster are placed together for better locality

### TODO
- [ ] use lookup table for distance computation in PQ search


crates/service/src/utils/cells.rs
```diff
@@ -23,4 +23,8 @@ impl<T: ?Sized> SyncUnsafeCell<T> {
     pub fn get_mut(&mut self) -> &mut T {
         self.value.get_mut()
     }
+
+    pub fn get_ref(&self) -> &T {
```
usamoi, line 27: Do not expose this function.

crates/service/src/algorithms/clustering/elkan_k_means.rs
```diff
@@ -24,30 +30,36 @@ impl<S: G> ElkanKMeans<S> {
...
         for i in 0..c {
             let mut sum = F32::zero();
-            for j in 0..n {
+            (0..n).into_par_iter().for_each(|j| {
                 let dis = S::elkan_k_means_distance(&samples[j], &centroids[i]);
-                lowerbound[(j, i)] = dis;
-                if dis * dis < weight[j] {
-                    weight[j] = dis * dis;
+                unsafe {
+                    (&mut *lowerbound.get())[(j, i)] = dis;
```
usamoi, line 45: It's absolutely unsound. Please use a structure that is like `Square<Atomic<F32>>`.

crates/service/src/algorithms/ivf/ivf_naive.rs
```diff
@@ -183,75 +176,120 @@ pub fn make<S: G>(
...
-        loop {
-            let next = heads[centroid_id as usize].load(Acquire);
-            unsafe {
-                nexts[i as usize].get().write(next);
-            }
-            let o = &heads[centroid_id as usize];
-            if o.compare_exchange(next, i, Release, Relaxed).is_ok() {
-                break;
+        unsafe {
+            (&mut *idx.get())[i as usize] = result.1 as usize;
```
usamoi, line 180: Unsound. Use `idx: Vec<AtomicUsize>` instead.

whateveraname:
### Performance Benchmark
Dataset: gist-960-euclidean-l2, n = 1,000,000, d = 960
|                         | build time (s) | rps | precision |
| ---------        | ---------- | ---- | --------- |
| IVF-naive       | 246         | 27.5 | 0.904      |
| IVF-naive-opt| 215         | 35.3 | 0.904      |
| IVF-PQ           | 3641       | 3.1   | 0.23         | 
| IVF-PQ-opt    | 1560      | 3.1    | 0.906      |

\*  Indices with 'opt' are optimized indices in this PR, num_threads in index build stage is set to 96
**   All indices use the default build parameters and select the search parameter (nprobe) to reach 0.9 precision
***  IVF-PQ has low precision of 0.23 due to the bug that PQ is not trained with residual vectors

whateveraname:
PTAL @usamoi

VoVAllen:
Why does num_threads=96 seem to have little acceleration? Is it due to the kmeans computation not parallized?

whateveraname:
> Thanks. Can you fix the lint error in CI? Also what's the PQ ratio in your benchmark?

The CI error is cause by the unused struct SyncUnsafeCell which is not written by me. I cannot decide whether to remove it. The PQ ratio is x4.

whateveraname:
> Why does num_threads=96 seem to have little acceleration? Is it due to the kmeans computation not paralleled?

It seems to me that a lot of time is spent on I/O so the speed up for computation cannot contribute much to overall performance.

VoVAllen:
> > Why does num_threads=96 seem to have little acceleration? Is it due to the kmeans computation not paralleled?
> 
> It seems to me that a lot of time is spent on I/O so the speed up for computation cannot contribute much to overall performance.

Doesn't make sense. 1M 960-dim vectors should be easy to fit in the memory. 

Also can you try benchmarking with higher PQ ratio for the results?

usamoi:
> > Thanks. Can you fix the lint error in CI? Also what's the PQ ratio in your benchmark?
> 
> The CI error is cause by the unused struct SyncUnsafeCell which is not written by me. I cannot decide whether to remove it. The PQ ratio is x4.

You can just remove unused code.

whateveraname:
> > > Why does num_threads=96 seem to have little acceleration? Is it due to the kmeans computation not paralleled?
> > 
> > 
> > It seems to me that a lot of time is spent on I/O so the speed up for computation cannot contribute much to overall performance.
> 
> Doesn't make sense. 1M 960-dim vectors should be easy to fit in the memory.
> 
> Also can you try benchmarking with higher PQ ratio for the results?

The index build time includes time to read the whole dataset from disk and time to save the whole index to disk, so it's not about whether vectors can fit in memory.

I will try for higher PQ ratio.

VoVAllen:
> > > > Why does num_threads=96 seem to have little acceleration? Is it due to the kmeans computation not paralleled?
> > > 
> > > 
> > > It seems to me that a lot of time is spent on I/O so the speed up for computation cannot contribute much to overall performance.
> > 
> > 
> > Doesn't make sense. 1M 960-dim vectors should be easy to fit in the memory.
> > Also can you try benchmarking with higher PQ ratio for the results?
> 
> The index build time includes time to read the whole dataset from disk and time to save the whole index to disk, so it's not about whether vectors can fit in memory.
> 
> I will try for higher PQ ratio.


It still doesn't make sense. Normal disk can achieve >500MB/s throughput for sequential write or read. 1M 960dim float is less than 4GB in space, which only accounts for less than 10s to read or write

VoVAllen (commented):
Also can you add some comments in the Kmeans and ivf part? Give a simple explanation on each part. And feel free to turn this PR in to ready mode, so we can start review

whateveraname:
> ### Performance Benchmark
> Dataset: gist-960-euclidean-l2, n = 1,000,000, d = 960
> 
> build time (s)	rps	precision
> IVF-naive	246	27.5	0.904
> IVF-naive-opt	215	35.3	0.904
> IVF-PQ-x4	3641	3.1	0.23
> IVF-PQ-x4-opt	368	3.1	0.906
> IVF-PQ-x16-opt	267	8.0	0.719
> * Indices with 'opt' are optimized indices in this PR, num_threads in index build stage is set to 96 ** All indices use the default build parameters and select the search parameter (nprobe) to reach 0.9 precision *** IVF-PQ has low precision of 0.23 due to the bug that PQ is not trained with residual vectors

### Update Benchmark

- Update build time for IVF-PQ-x4-opt run with the latest commit, which parallelizes over training for subquantizers
- Add result for IVF-PQ-x16-opt

VoVAllen:
PTAL @usamoi

crates/service/src/algorithms/quantization/product.rs
```diff
@@ -69,12 +72,16 @@ unsafe impl<S: G> Send for ProductQuantization<S> {}
...
 impl<S: G> ProductQuantization<S> {
-    fn codes(&self, i: u32) -> &[u8] {
+    pub fn codes(&self, i: u32) -> &[u8] {
         let width = self.dims.div_ceil(self.ratio);
         let s = i as usize * width as usize;
         let e = (i + 1) as usize * width as usize;
         &self.codes[s..e]
     }
+
+    pub fn set_codes(&mut self, codes: MmapArray<u8>) {
```
usamoi, line 82: Do not expose it.

whateveraname: fixed, PTAL

VoVAllen:
Please fix the CI check

whateveraname:
> The IVFPQ uses table lookup for distance computation in the search stage now. Currently only supports L2 distance. Will add support for IP distance next. Cosine distance will remain using the original distance computation method. The following table is the performance benchmark.
> 
> build time (s)	rps	precision
> IVF-PQ-x4-opt	368	3.08	0.906
> IVF-PQ-x4-opt-table	341	5.27	0.901
> IVF-PQ-x16-opt	267	8.05	0.719
> IVF-PQ-x16-opt-table	251	19.26	0.720

 
During IVFPQ search with by_residual, we compute
 &emsp; d = || x - y_C - y_R ||^2
where x is the query vector, y_C the coarse centroid, y_R the refined PQ centroid. The expression can be decomposed as:
 &emsp;       d = || x - y_C ||^2 + || y_R ||^2 + 2 * (y_C | y_R) - 2 * (x | y_R)
&emsp;&emsp;&emsp;           --------- &emsp; &emsp; -------------------------- &ensp;      -----------
&emsp;&emsp;&emsp;&ensp;               term 1       &emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;          term 2                   &emsp;&emsp;&emsp;&emsp;&emsp;&emsp;term 3
When using multiprobe, we use the following decomposition:
- term 1 is the distance to the coarse centroid, that is computed during the 1st stage search.
- term 2 can be precomputed, as it does not involve x.
- term 3 is the classical non-residual distance table. Since y_R defined by a product quantizer, it is split across subvectors and stored separately for each subvector.

At search time, the tables for term 2 and term 3 are added up. This is faster when the length of the lists is > ksub * M.

usamoi:
Is it ready for merging?

whateveraname:
> Is it ready for merging?

It is ready for merging now

usamoi (approved):

Issue #327, opened 2024-01-31
feat: Support vector prefix chunk when indexing


VoVAllen:
The latest OpenAI embedding support using subvector as dimension reduction. User should be able to index the vectors based on that like 
```SQL
create index on items using vectors (prefix_chunk(embedding, 64) vector_l2_ops) 
```

Reference: https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions

usamoi:
I think it can be supported by creating an index on expression.

usamoi:
Completed in #329 

Should we document it now? (We do not release documentation with software so users may read text about a feature that is not implemented in `2.0`)

usamoi:
Document is tracking in #255 and #358

gaocegege:
How could I use this feature? Do we have a pre-release for it? I am writing the benchmark for it.

VoVAllen:
@gaocegege `CREATE INDEX t_vector_index on t using vectors((val[0:128]::vector(128)) vector_l2_ops);` to create index on prefix 128 dim. I think the doc is missing now 😢

gaocegege:
Would you mind creating a pre-release for test purpose?

VoVAllen:
Let me check how I can do this. Probably to include #368 also. And I just realized the nightly docker image doesn't work now.

gaocegege:
OK, then let's wait for #368

usamoi:
When will we fix nightly releases?

VoVAllen:
tracked in https://github.com/tensorchord/pgvecto.rs/issues/373

PR #329, opened 2024-02-01
feat: support vector subscript

usamoi:
Address #327

Now we support such a syntax:

```sql
CREATE TABLE t(val vector(3));
INSERT INTO t (val) SELECT ARRAY[random(), random(), random()]::real[] FROM generate_series(1, 100000);
CREATE INDEX t_vector_index on t using vectors((val[:1]::vector(1)) vector_l2_ops);
SELECT * FROM t ORDER BY val[:1]::vector(1) <-> '[0]' limit 10;
```

It's only supported by pg14, pg15 and pg16.


VoVAllen:
The syntax using real[] is not consistent with other usage with vectors. Is it necessary?

usamoi:
> The syntax using real[] is not consistent with other usage with vectors.

It's nothing about `real[]`.

> Is `vectors` a new function?

It's `vectors` AM handler. You can get reference of syntax in https://www.postgresql.org/docs/current/sql-createindex.html.

VoVAllen:
Can I do `CREATE INDEX t_vector_index on t using vectors((val[16:32]::vector(16)) vector_l2_ops);` ?

usamoi:
> Can I do `CREATE INDEX t_vector_index on t using vectors((val[16:32]::vector(16)) vector_l2_ops);` ?

Yes.

VoVAllen (dismissed):

PR #339, opened 2024-02-02
fix: remove ctor

usamoi:
For reference: https://github.com/mmastrac/rust-ctor/issues/280

The simplest solution is to just remove it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,16 +440,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctor"
-version = "0.2.6"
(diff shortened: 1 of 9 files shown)


VoVAllen:
What's ctor for? Does it cause any bug in the current version?

usamoi:
> What's ctor for? Does it cause any bug in the current version?

It's for initializing instruction detection so that we do not need to check `cpuid` every time.

`rustc` does not add `--whole-archive` to linker options for deps so it's possible that functions marked `ctor` do not actually run.

In current binary releases it does not cause any bugs but we cannot guarantee it.

PR #342, opened 2024-02-05
chore: use latest commit SHA to trigger the nightly release

kemingy:


kemingy:
It seems the release edit type only checks the **release title**.

I have verified that the `gh release upload` will override the old artifacts.

gaocegege (approved):

PR #350, opened 2024-02-07
feat: new openai embedding

cutecutecat:


src/gucs/embedding.rs
```diff
@@ -0,0 +1,94 @@
...
+        .to_str()
+        .map_err(|_| SessionError::GucParseFailed {
+            key: name,
+            hint: "utf8 parse failed".to_string(),
+        })?
+        .to_string();
+    Ok(ret)
+}
+
+pub fn openai_options() -> Result<OpenAIOptions, SessionError> {
```
usamoi, line 33: What about using `JSON` as second argument of `text2vec_openai`?

src/embedding/udf.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::gucs::embedding::{openai_options, OpenAIOptions};
+use crate::prelude::embedding_failed;
+use pgrx::default;
+use reqwest::blocking::Client;
+use service::prelude::F32;
+
+use super::openai::{EmbeddingRequest, EmbeddingResponse};
+
+#[pgrx::pg_extern(volatile, strict)]
+fn _vectors_ai_embedding_vector(
```
usamoi, line 13: Why we do not expose it in `finialize.sql`?

src/embedding/udf.rs
```diff
@@ -0,0 +1,143 @@
...
+    input: String,
+    model: default!(String, "'text-embedding-ada-002'"),
```
usamoi, line 15: A model may be not popular or depcrated. It's not very good to set default value for `model`.

src/gucs/embedding.rs
```diff
@@ -0,0 +1,54 @@
...
+    let api_key = guc_string_parse(&OPENAI_API_KEY, "vectors.openai_api_key".to_string());
+    OpenAIOptions { base_url, api_key }
+}
+
+static OPENAI_API_KEY: GucSetting<Option<&'static CStr>> =
+    GucSetting::<Option<&'static CStr>>::new(None);
+
+static OPENAI_BASE_URL: GucSetting<Option<&'static CStr>> =
+    GucSetting::<Option<&'static CStr>>::new(Some(unsafe {
+        CStr::from_bytes_with_nul_unchecked(b"https://api.openai.com/v1/\0")
```
usamoi, line 34: `c_str_literals` is stablized in latest nightly, and we so update toolchain and use `CStr` literals.

src/gucs/embedding.rs
```diff
@@ -0,0 +1,54 @@
...
+        GucFlags::default(),
+    );
+    GucRegistry::define_string_guc(
+        "vectors.openai_base",
+        "The base url of OpenAI or compatible server.",
+        "",
+        &OPENAI_BASE_URL,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
```
usamoi, line 53:
```suggestion
    GucRegistry::define_string_guc(
        "vectors.openai_base_url",
        "The base url of OpenAI or compatible server.",
        "",
        &OPENAI_BASE_URL,
        GucContext::Userset,
        GucFlags::default(),
    );
```

src/prelude/error.rs
```diff
@@ -11,6 +15,22 @@ You should edit `shared_preload_libraries` in `postgresql.conf` to include `vect
...
 
+pub fn embedding_failed(hint: &str) -> ! {
+    error!(
+        "\
+Error happens at embedding.
+INFORMATION: hint = {hint}"
+    );
+}
+
+pub fn guc_parse_failed(key: &str, hint: &str) -> ! {
```
usamoi, line 26:
```suggestion
pub fn bad_guc_literal(key: &str, hint: &str) -> ! {
```

src/prelude/error.rs
```diff
@@ -11,6 +15,22 @@ You should edit `shared_preload_libraries` in `postgresql.conf` to include `vect
 or simply run the command `psql -U postgres -c 'ALTER SYSTEM SET shared_preload_libraries = \"vectors.so\"'`.");
 }
 
+pub fn embedding_failed(hint: &str) -> ! {
```
usamoi, line 18:
```suggestion
pub fn bad_embedding(hint: &str) -> ! {
```

src/gucs/embedding.rs
```diff
@@ -0,0 +1,54 @@
+use std::ffi::CStr;
+
+use pgrx::{GucContext, GucFlags, GucRegistry, GucSetting};
+
+use crate::prelude::guc_parse_failed;
+
+fn guc_string_parse(target: &'static GucSetting<Option<&'static CStr>>, name: String) -> String {
```
usamoi, line 7:
```suggestion
fn guc_string_parse(target: &'static GucSetting<Option<&'static CStr>>, name: &'static str) -> String {
```

src/gucs/embedding.rs
```diff
@@ -0,0 +1,54 @@
...
+    let value = match target.get() {
+        Some(s) => s,
+        None => guc_parse_failed(&name, "uninitialized"),
```
usamoi, line 10:
```suggestion
        None => guc_parse_failed(&name, "should not be `NULL`"),
```

src/gucs/embedding.rs
```diff
@@ -0,0 +1,54 @@
...
+    };
+    match value.to_str() {
+        Ok(s) => s.to_string(),
+        Err(_e) => guc_parse_failed(&name, "utf8 parse failed"),
```
usamoi, line 14:
```suggestion
        Err(_e) => guc_parse_failed(&name, "should be a valid UTF-8 string"),
```

src/gucs/embedding.rs
usamoi, line 7: We can take advantage of a checker functions for every string GUC to ensure it's valid UTF-8.

src/prelude/error.rs
```diff
@@ -11,6 +15,22 @@ You should edit `shared_preload_libraries` in `postgresql.conf` to include `vect
...
+Error happens at embedding.
+INFORMATION: hint = {hint}"
+    );
+}
+
+pub fn guc_parse_failed(key: &str, hint: &str) -> ! {
+    error!(
+        "\
+        Failed to parse Guc.
+        INFORMATION: gucs = {key}, hint = {hint}"
```
usamoi, line 30:
```suggestion
        INFORMATION: GUC = {key}, hint = {hint}"
```

src/prelude/error.rs
```diff
@@ -11,6 +15,22 @@ You should edit `shared_preload_libraries` in `postgresql.conf` to include `vect
...
+        "\
+Error happens at embedding.
+INFORMATION: hint = {hint}"
+    );
+}
+
+pub fn guc_parse_failed(key: &str, hint: &str) -> ! {
+    error!(
+        "\
+        Failed to parse Guc.
```
usamoi, line 29:
```suggestion
        Failed to parse a GUC variable.
```

src/prelude/error.rs
```diff
@@ -1,7 +1,11 @@
 use crate::ipc::{ClientRpc, ConnectionError};
-use pgrx::error;
 use std::num::NonZeroU16;
 
+#[cfg(not(test))]
+use pgrx::error;
+#[cfg(test)]
+use std::panic as error;
```
usamoi, line 7: We'd better move most of these code to a separate crate and make these functions to return errors instead of just panicking.

src/embedding/udf.rs
cutecutecat: This is widely used as default: https://github.com/search?q=text-embedding-ada-002&type=code
As a stright string, it easy to maintain when it is depcrated, and it's important for user to input as less as he could.

VoVAllen: `text-embedding-3-small` should be the new default. It's 5x cheaper than 002 and has better performance

usamoi: Someone who uses model `abc` could always create a new function named `text2vec_openai_abc` or `my_text2vec` to call `text2vec_openai`.

usamoi: We can add `text2vec_openai_ada002` and `text2vec_openai_3small`.

VoVAllen: > We can add `text2vec_openai_ada002` and `text2vec_openai_3small`. (I think it is a user's duty but it's not bad if we provide it.)

LGTM. The problem of default model name is that there are many embedding providers nowadays such as jina/cohere/voyage. They have different model names.

src/embedding/udf.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::datatype::vecf32::{Vecf32, Vecf32Output};
+use crate::gucs::embedding::{openai_options, OpenAIOptions};
+use pgrx::error;
+use reqwest::blocking::Client;
+use service::prelude::F32;
+
+use super::openai::{EmbeddingError, EmbeddingRequest, EmbeddingResponse};
+
+#[pgrx::pg_extern(volatile, strict)]
+fn _vectors_ai_embedding_vector_v3(input: String) -> Vecf32Output {
```
usamoi, line 12: Writing this function in PL/pgSQL might be better.

src/embedding/openai.rs
usamoi, line 1: Move this file to `crates/embedding/src` since we can use and test it independently.

src/sql/finalize.sql
```diff
@@ -340,6 +340,12 @@ CREATE OPERATOR <=> (
 
 -- List of functions
 
+CREATE FUNCTION text2vec_openai(input TEXT, model TEXT) RETURNS vector
+STRICT LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_ai_embedding_vector_wrapper';
+
+CREATE FUNCTION text2vec_openai_v3(input TEXT) RETURNS vector
```
usamoi, line 346: There is also `text-embedding-3-large` so `3small` here is better. Let's also add `3large` and `ada002`.

src/sql/finalize.sql
```diff
@@ -340,6 +340,12 @@ CREATE OPERATOR <=> (
 
 -- List of functions
 
+CREATE FUNCTION text2vec_openai(input TEXT, model TEXT) RETURNS vector
```
usamoi, line 343: Put these function in the end of `List of functions`.

src/gucs/mod.rs
```diff
@@ -7,5 +12,20 @@ pub unsafe fn init() {
         self::planning::init();
         self::internal::init();
         self::executing::init();
+        self::embedding::init();
+    }
+}
+
+fn guc_string_parse(
```
usamoi, line 19: There is a GUC check hook to check if the string is valid UTF-8 string. Let's use that.

src/sql/finalize.sql
cutecutecat: It's a substitute of default value, we only need to write the most modern one, and other could let user to pick `text2vec_openai`.

If we encode all models, there would be a huge amount of functions in the future.

src/sql/finalize.sql
```diff
@@ -346,6 +346,20 @@ STRICT LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_pgvectors_upgrade_wrapper';
 CREATE FUNCTION to_svector(dims INT, indices INT[], vals real[]) RETURNS svector
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_to_svector_wrapper';
 
+CREATE FUNCTION text2vec_openai(input TEXT, model TEXT) RETURNS vector
+STRICT LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_text2vec_openai_wrapper';
+
+CREATE FUNCTION text2vec_openai_v3(input TEXT) RETURNS vector
```
usamoi, line 352: I still thinks `text2vec_openai_3small` is better. God knows if OpenAI will release `text-embedding-3-turbo-small` or `text-embedding-3-very-small`.

usamoi (approved):

src/sql/finalize.sql
cutecutecat: We could remind and track it when further update.

PR #368, opened 2024-02-20
feat: support binary vector

### LINKED ISSUE #355: feat: Binary vector support
Support binary vector type and user can use it to build the index.

[Benchmark](https://twitter.com/NirantK/status/1755135939901628877/photo/1) result from Qdrant:
image

silver-ymz:
close #355


silver-ymz:
PTAL @VoVAllen @usamoi

VoVAllen:
What's the alignment guarantee of bitvec? And I noticed it said SIMD is not implemented. We might need to implement the simd part later. Shall we define our own Bitvec? How big is the workload for this?

usamoi:
I'm working on refactoring so this PR may not be merged this week.

VoVAllen:
@usamoi We need this feature to do some benchmark with OpenAI embedding. If not going to merge this week, can you or @silver-ymz help build a Docker image for this branch?

VoVAllen:
@silver-ymz You can try the following things in remaining time this week:
- Try to build a Docker image on this branch and push it to your own Docker Hub account.
- SIMD acceleration for bit operations

silver-ymz:
> What's the alignment guarantee of bitvec?

It's aligned to `usize`. `bitslice` save an unaligned pointer to split a slice into three parts, header, body and tail. In our implementation, we ensures that every time constructing `BinaryVec`, there is no header bits in order to copy freely.

> And I noticed it said SIMD is not implemented. We might need to implement the simd part later. Shall we define our own Bitvec? How big is the workload for this?

I think as long as casting `bitslice` to `&[usize]`, it can take advantage of auto vectorization. Maybe it won't take a lot of time.

VoVAllen:
https://github.com/ferrilab/bitvec/blob/5fb855073acc2ed045094ed89d8daf8c765f0135/src/slice.rs#L1062-L1075
It doesn't look like can be auto vectorized.

crates/base/src/vector/bvecf32.rs
```diff
@@ -0,0 +1,153 @@
+use super::{VectorBorrowed, VectorOwned};
+use crate::scalar::F32;
+use serde::{Deserialize, Serialize};
+
+pub const BVEC_WIDTH: usize = std::mem::size_of::<usize>() * 8;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BVecf32Owned {
```
usamoi, line 8: We can add a new scalar type `bit` (boolean).

crates/base/src/global/bvecf32.rs
usamoi, line 1: For binary vectors, users may need Hamming distance or Jaccard distance.

crates/base/src/vector/bvecf32.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+    #[inline(always)]
+    pub fn new_checked(dims: u16, data: &'a [usize]) -> Option<Self> {
+        if dims == 0 {
+            return None;
+        }
+        if data.len() != (dims as usize).div_ceil(BVEC_WIDTH) {
+            return None;
+        }
+        unsafe { Some(Self::new_unchecked(dims, data)) }
```
usamoi, line 78: We should check the padding bits are zero.

crates/c/src/binary.c
```diff
@@ -0,0 +1,209 @@
...
+
+#if defined(__x86_64__)
+#include <immintrin.h>
+#endif
+
+#if defined(__x86_64__)
+
+#define WIDTH (512 / 8 / sizeof(size_t))
+
+__attribute__((target("arch=x86-64-v4,avx512vpopcntdq"))) extern float
```
usamoi, line 12: `avx512vpopcntdq` intrinsics are implemented in Rust so it is not necessary to write the code in C.

src/datatype/memory_bvecf32.rs
```diff
@@ -0,0 +1,168 @@
...
+use std::ops::Deref;
+use std::ops::DerefMut;
+use std::ptr::NonNull;
+
+#[repr(C, align(8))]
+pub struct BVecf32Header {
+    varlena: u32,
+    dims: u16,
+    kind: u8,
+    reserved: u8,
```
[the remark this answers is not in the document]
silver-ymz: completed in new commit

crates/base/src/vector/bvecf32.rs
```diff
@@ -0,0 +1,153 @@
...
+        assert!(index < self.dims as usize);
+        if value {
+            self.data[index / BVEC_WIDTH] |= 1 << (index % BVEC_WIDTH);
+        } else {
+            self.data[index / BVEC_WIDTH] &= !(1 << (index % BVEC_WIDTH));
+        }
+    }
+
+    #[inline(always)]
+    pub fn data_mut(&mut self) -> &mut [usize] {
```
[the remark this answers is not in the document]
silver-ymz: I set it to `unsafe`.

crates/base/src/global/bvecf32.rs
silver-ymz: Jaccard distance completed in new commit. It seems that for two equal-length vectors, Hamming distance equals to squared Euclidean distance. We can only add some documents.

crates/c/src/binary.c
silver-ymz: completed in new commit

crates/base/src/vector/bvecf32.rs
silver-ymz: I don't understand. Where can we add a new scalar type `bit`?

silver-ymz: But `+` and `-` have different meaning for `bit` and `float`. `bit` seems not unsuitable for `scalar` abstraction.

crates/base/src/global/bvecf32.rs
usamoi: Users may want to create indexes on Jaccard distance but not on cosine/dot distance. Milvus supports both Jaccard and Hamming and faiss supports Hamming only.

src/datatype/casts.rs
```diff
@@ -83,3 +84,35 @@ fn _vectors_cast_svecf32_to_vecf32(
...
+fn _vectors_cast_vecf32_to_bvecf32(
+    vector: Vecf32Input<'_>,
+    _typmod: i32,
+    _explicit: bool,
+) -> BVecf32Output {
+    let mut values = BVecf32Owned::new_zeroed(vector.len() as u16);
+    for (i, &x) in vector.slice().iter().enumerate() {
+        match x.to_f32() {
+            x if x == 0. => {}
+            x if x == 1. => values.set(i, true),
```
VoVAllen, line 98: Shall we change it to `1 if  x>0`? @usamoi

usamoi: If we want to accept errors, we could use `EPS`. But I think it's not necessary here.

VoVAllen: Currently only [0,1,1] vector works. But we need the casting to do the binary quantization. I personally think it's fine to let element bigger than zero to be 1, and other to be 0. What do you think?

usamoi: We can provide a function to do it. It's more explicit.

gaocegege: And should not be == 0. There are some values below 0.

e.g. [ -0.022388368844985962, 0.028537113219499588, -0.017491688951849937, -0.01533466950058937, -0.03451230376958847, 0.0162644200026989, -0.020231351256370544, -0.012743767350912094, -0.032082557678222656, 0.0037468906957656145, -0.003864658996462822, 0.004292343743145466, 0.007035105023533106, -0.03314867243170738, ...]

crates/base/src/global/bvecf32.rs
silver-ymz: completed in new commit

src/datatype/casts.rs
silver-ymz: completed in new commit

usamoi (approved):

PR #395, opened 2024-02-29
chore: upgrade 0.2.1 script

### LINKED ISSUE #380: Issue upgrade from 0.1.10 (thru 0.1.11) to 0.2.0
The upgrade hit https://github.com/tensorchord/pgvecto.rs/blob/main/sql/upgrade/vectors--0.1.10--0.1.11.sql#L23-L27 which would result in `ERROR: could not find function "vector_stat_wrapper" in file "/usr/lib/postgresql/16/lib/vectors.so"`. This is especially an issue when I'm using the docker image.

I ended up manually overwritten the `sql/upgrade/vectors--0.1.10--0.1.11.sql` commenting out https://github.com/tensorchord/pgvecto.rs/blob/main/sql/upgrade/vectors--0.1.10--0.1.11.sql#L23-L40
```
CREATE FUNCTION "vector_stat"(
...
```
as well as overwritten `sql/upgrade/vectors--0.1.11--0.2.0.sql` commenting out https://github.com/tensorchord/pgvecto.rs/blob/main/sql/upgrade/vectors--0.1.11--0.2.0.sql#L193 & https://github.com/tensorchord/pgvecto.rs/blob/main/sql/upgrade/vectors--0.1.11--0.2.0.sql#L463
`DROP VIEW pg_vector_index_info;` & `DROP FUNCTION vector_stat;`

cutecutecat:
Close #380

- Add apt update to release script, sometimes it would fail to install
- Add install script for v0.2.1
- Add upgrade route 0.1.0 -> 0.2.0
- Add upgrade route 0.2.0 -> 0.2.1
- Delete upgrade route 0.1.0 -> 0.1.1 to avoid 0.1.0->0.1.1->0.2.0 upgrade
- Fix typo

The cherry pick would be later


sql/upgrade/vectors--0.1.10--0.1.11.sql
usamoi, line 1: Why remove it?

https://www.postgresql.org/docs/current/extend-extensions.html

> PostgreSQL doesn't assume anything about the properties of version names: for example, it does not know whether 1.1 follows 1.0. It just matches up the available version names and follows the path that requires applying the fewest update scripts. (A version name can actually be any string that doesn't contain -- or leading or trailing -.)

sql/upgrade/vectors--0.2.0--0.2.1.sql
```diff
@@ -0,0 +1,10 @@
+\echo Use "ALTER EXTENSION vectors UPDATE TO '0.2.1'" to load this file. \quit
+
+DO LANGUAGE plpgsql $$
+    DECLARE
+    BEGIN
+        IF '@extschema@' != 'vectors' THEN
+            RAISE EXCEPTION 'Please read upgrade document in https://docs.pgvecto.rs/admin/upgrading.html#upgrade-from-0-1-x.';
+        END IF;
+    END;
+$$;
```
usamoi, line 10: We do not need to add this hint for target versions `> 0.2.0`.

sql/upgrade/vectors--0.1.12--0.1.13.sql
```diff
@@ -1,4 +1,4 @@
-\echo Use "ALTER EXTENSION vectors UPDATE TO '0.2.0'" to load this file. \quit
+\echo Use "ALTER EXTENSION vectors UPDATE TO '0.1.13'" to load this file. \quit
```
usamoi, line 1: It's not a typo. We do not really support update among versions `< 0.2.0` since update scripts are not included in these releases.

PR #399, opened 2024-03-01
feat: support sparse vector with 1_048_575 dims

### LINKED ISSUE #384: feat: use int32 for sparse vector index

silver-ymz:
close #384


silver-ymz:
PTAL @usamoi @VoVAllen

usamoi:
Can you replace `1000000` with `1048576`? It's hard to count there are how many zeros.

silver-ymz:
> Could you replace 1000000 with 1048575 ($2^{20} - 1$)?

updated

crates/service/src/algorithms/quantization/product.rs
```diff
@@ -67,7 +67,7 @@ impl<S: G> Quan<S> for ProductQuantization<S> {
         let precomputed_table =
             serde_json::from_slice(&std::fs::read(path.join("table")).unwrap()).unwrap();
         Self {
-            dims: options.vector.dims,
+            dims: options.vector.dims.try_into().unwrap(),
```
usamoi, line 70: Let's alter `dims: u16` to `dims: u32` too.

src/datatype/text_bvecf32.rs
```diff
@@ -13,28 +13,27 @@ fn _vectors_bvecf32_in(input: &CStr, _oid: Oid, typmod: i32) -> BVecf32Output {
...
-        s.parse::<u8>().ok().and_then(|x| match x {
-            0 => Some(false),
-            1 => Some(true),
-            _ => None,
-        })
-    });
-    match v {
-        Err(e) => {
-            bad_literal(&e.to_string());
+    let mut bool_vec = Vec::<bool>::with_capacity(reserve as usize);
```
usamoi, line 16: Why?

src/datatype/text_svecf32.rs
```diff
@@ -8,75 +8,26 @@ use std::ffi::{CStr, CString};
...
-                let value: F32 = solve(token.parse().ok(), "Bad number.");
-                if !value.is_zero() {
-                    indexes.push(index);
-                    values.push(value);
+    if let Err(e) = parse_vector(input.to_bytes(), |i, s| {
+        s.parse::<F32>()
+            .ok()
+            .map(|s| {
+                dims = i + 1;
+                if !s.is_zero() {
```
usamoi, line 20: It's not a good idea to pass a function that produces side-effects to `map`.

src/datatype/typmod.rs
```diff
@@ -50,10 +50,24 @@ fn _vectors_typmod_in(list: Array<&CStr>) -> i32 {
...
         typmod.into_i32()
     } else {
-        check_type_dims(None);
+        check_type_dims_u16(None);
+        unreachable!()
+    }
+}
+
+#[pgrx::pg_extern(immutable, parallel_safe, strict)]
+fn _vectors_typmod_in_i32(list: Array<&CStr>) -> i32 {
```
usamoi, line 62:
```suggestion
fn _vectors_typmod_in_1048575(list: Array<&CStr>) -> i32 {
```

src/ipc/transport/mmap.rs
```diff
@@ -81,6 +81,9 @@ impl Socket {
         }
     }
     pub fn send(&mut self, packet: &[u8]) -> Result<(), ConnectionError> {
+        if packet.len() > BUFFER_SIZE - 8 {
```
usamoi, line 84: `Bincode` will emit the length of `packet` to bytestream so `BUFFER_SIZE - 8` is not a strict upperbound. Let's check it when serializing.

src/datatype/text_bvecf32.rs
silver-ymz: I want to use `parse_vector` for `svector`. Therefore, I changed `parse_vector` to pass all the strs that need to be parsed to a hook function instead of returning a vector.

crates/service/src/algorithms/quantization/product.rs
silver-ymz: completed in new commit

src/datatype/typmod.rs
silver-ymz: completed in new commit

src/datatype/text_svecf32.rs
silver-ymz: I replace all `map` with side-effects to pattern match.

src/datatype/text_bvecf32.rs
usamoi: We can just let `parse_vector` for `svector` return `Vec<F32>` too. Allocation is not a big ideal.

silver-ymz: update `parse_vector` to previous impl in new commit

usamoi (approved):

PR #401, opened 2024-03-01
fix: bvector subscription may modify padding bits

silver-ymz:


silver-ymz:
PTAL @usamoi @VoVAllen

usamoi (approved):

PR #416, opened 2024-03-11
fix: grant permission for `pg_vector_index_stat` in install sql

### LINKED ISSUE #412: chore: Fix permission for pg_vector_index_stat
Should grant permission to the view when install the extension

Reference issue: https://github.com/immich-app/immich/issues/7626

image
image

usamoi:
* closes #412

### DIFF
--- a/src/sql/finalize.sql
+++ b/src/sql/finalize.sql
@@ -742,4 +742,6 @@ CREATE VIEW pg_vector_index_stat AS
          pg_am A ON A.oid = I.relam
     WHERE A.amname = 'vectors';
 
+GRANT SELECT ON TABLE pg_vector_index_stat TO PUBLIC;
+
 -- finalize end


usamoi:
Users can do it by themselves if they meet issues.

________________________________

Ornias1993:
> Please cherry pick this into 0.2 branch

This is not a full non-beta release product. It's pretty much hot-but-good code, when implementing hot-new versions, you should do a lot of testing to ensure bugs like these don't hit your codebase. That's the burden of using pre-release projects like Vecto.rs.
They cannot send-out patch releases for every single bugfix.

Best blame Immich, i'm betting that's the source of your troubles, for overzealusly implementing pgvecto.rs without thorough superuser-and-non-superuser testing.

VoVAllen:
@Ornias1993 Yeah, The diversity of user deployment methods for immich is something we hadn't met before. And the technique behind context search with filter is based on a new paper VBASE published in 2023. That's also the reason why we looks kind of in the beta status.

Luckily (or a bit sadly), all the complains so far are about the upgrade/installation problems, but not the function itself. We'll work closely with immich team to improve the user experiences.

PR #419, opened 2024-03-13
feat: Support binary copy in python SDK.

### LINKED ISSUE #354: feat: Support binary representation in python SDK
Inserting data in binary representation significantly accelerates performance in PostgreSQL. But it is hard to write in python. We should support it in SDK

Could you please put the sample python code in the issue? @VoVAllen

my-vegetable-has-exploded:
close #354


my-vegetable-has-exploded:
The python tests works well on my machine,  but falls in ci.  I don't figure out the reason.
图片
And for sparse vector,  I think we need a type to descript it.  Sparse vector's dimension is not the same with length, so we can't get it from ndarray.type and the adapter need to a type to register.
图片

my-vegetable-has-exploded:
It seems that our workflow use v0.2.1 not nightly image. I think ci will be happy after using nightly.

bindings/python/examples/perf.sh
```diff
@@ -0,0 +1,5 @@
+/bin/python3 /home/envd/pgvecto.rs/bindings/python/examples/psycopg_testcpy.py &
```
my-vegetable-has-exploded, line 1: Sorry, I would remove this shell later.

my-vegetable-has-exploded:
Thanks!

VoVAllen:
@all-contributors please add @my-vegetable-has-exploded for code contributions

Issue #250, opened 2024-01-10
feat: support multicolumn index


usamoi:
support multicolumn index to allow users to take advantage of prefiltering

PR #429, opened 2024-03-17
feat: support multicolumn index

silver-ymz:
Part of #250

It's able to create a multicolumn index and save a `bigint` to payload.

Usage:
```sql
CREATE TABLE t (id serial, val vector(3));

INSERT INTO t (val) SELECT ARRAY[random(), random(), random()]::real[] FROM generate_series(1, 1000);

CREATE INDEX ON t USING vectors (val vector_l2_ops, (id::bigint))
WITH (options = "[indexing.flat]");
```

Limit:
It can only use 2 columns. The first column must be a vector type, and the second column must be `bigint`.

### DIFF
--- a/src/index/am.rs
+++ b/src/index/am.rs
@@ -51,7 +51,7 @@ const AM_HANDLER: pgrx::pg_sys::IndexAmRoutine = {
     am_routine.amcanorderbyop = true;
     am_routine.amcanbackward = false;
     am_routine.amcanunique = false;
-    am_routine.amcanmulticol = false;
+    am_routine.amcanmulticol = true;
     am_routine.amoptionalkey = true;
     am_routine.amsearcharray = false;
     am_routine.amsearchnulls = false;
@@ -207,8 +207,13 @@ pub unsafe extern "C" fn aminsert(
     let oid = (*index_relation).rd_id;
     let id = get_handle(oid);
     let vector = from_datum(*values.add(0), *is_null.add(0));
+    let multicolumn_data = if (*(*index_relation).rd_index).indnkeyatts == 2 {
+        pgrx::datum::FromDatum::from_datum(*values.add(1), *is_null.add(1)).unwrap()
+    } else {
+        0
+    };
     if let Some(v) = vector {
-        am_update::update_insert(id, v, *heap_tid);
+        am_update::update_insert(id, v, *heap_tid, multicolumn_data);
     }
     true
 }
--- a/src/sql/finalize.sql
+++ b/src/sql/finalize.sql
@@ -661,6 +661,8 @@ CREATE OPERATOR FAMILY veci8_dot_ops USING vectors;
 
 CREATE OPERATOR FAMILY veci8_cos_ops USING vectors;
 
+CREATE OPERATOR FAMILY bigint_attr_ops USING vectors;
+
 -- List of operator classes
 
 CREATE OPERATOR CLASS vector_l2_ops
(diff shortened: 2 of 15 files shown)


src/index/am.rs
```diff
@@ -207,8 +207,13 @@ pub unsafe extern "C" fn aminsert(
     let oid = (*index_relation).rd_id;
     let id = get_handle(oid);
     let vector = from_datum(*values.add(0), *is_null.add(0));
+    let multicolumn_data = if (*(*index_relation).rd_index).indnkeyatts == 2 {
+        pgrx::datum::FromDatum::from_datum(*values.add(1), *is_null.add(1)).unwrap()
```
usamoi, line 211: We must handle `NULL`s here.

> A related restriction is that an index access method that supports multiple index columns must support indexing null values in columns after the first, because the planner will assume the index can be used for queries that do not restrict these columns.

src/sql/finalize.sql
```diff
@@ -727,6 +729,10 @@ CREATE OPERATOR CLASS veci8_cos_ops
     FOR TYPE veci8 USING vectors AS
     OPERATOR 1 <=> (veci8, veci8) FOR ORDER BY float_ops;
 
+CREATE OPERATOR CLASS bigint_attr_ops
+    DEFAULT FOR TYPE bigint USING vectors AS
+    OPERATOR 1 = (bigint, bigint);
```
usamoi, line 734: We should add a function here to convert any datatypes to an integer.

silver-ymz: > If neither `FOR SEARCH` nor `FOR ORDER BY` is specified, `FOR SEARCH` is the default.

Maybe `FOR SEARCH` is better than `FOR ORDER BY`, because the results returned by the backend are not sorted by the second column.

usamoi: If we take advantage of `OPERATOR FOR SEARCH`, maybe we will lost some chances about optimization, for example:

* `SELECT * FROM t ORDER BY embedding <-> '[0, 0, 0]' WHERE label = 'alpha' OR label = 'beta'`. It's a classical HNSW query (see https://qdrant.tech/articles/filtrable-hnsw/, "Select points which belong to a specific subset of categories"). But `label = 'alpha' OR label = 'beta'` will be a filter cond, not an index clause. It can be solved if we provide a new operator `<:` and force users to write `label <: ('label', 'beta')`.
* `SELECT * FROM t ORDER BY embedding <-> '[0, 0, 0]' WHERE id = id`. `id = id` is also a filter cond, not an index clause. PostgreSQL will fetch rows in table. But if it's checked by an index, it's faster. It can be solved if we support index only scan...

VoVAllen: I did some investigation last week, on how btree/heap index filter about the where condition. It seems we can reuse this part of the code like https://github.com/postgres/postgres/blob/master/src/backend/access/heap/heapam.c#L792-L793. Is it necessary for us to define a new operator class here?

usamoi: > Is it necessary for us to define a new operator class here?

How if a new operator is defined here? `ambuild/aminsert` get column data, and shouldn't the index need to known it's an integer, string or something else?

VoVAllen: I'm not sure. But does btree or other postgres need a new operator type? I think the case here should be similar.

VoVAllen: GIN/GIST index also support data type like tsvector. How does they know the real data and push down the filter condition to the index scan?

VoVAllen:
The limitation is totally reasonable. I don't think we'll support more general cases here soon. Filter by category id/range query should be enough with number/int type. A more complex problem is to store product type as bitmap(or roaringbitmap). However we don't need to consider such complicated scenario now.

PR #444, opened 2024-03-22
fix: invalid check for bvector with 64 dims

silver-ymz:


silver-ymz:
PTAL @usamoi @VoVAllen

usamoi (approved):

PR #448, opened 2024-03-27
feat: stand alone test

whateveraname:
Test algorithms without pg. See demo test files in crates/stand-alone-test/src/bin.


crates/stand-alone-test/src/bin/search_hnsw.rs
```diff
@@ -0,0 +1,78 @@
...
+struct FilterStruct {}
+impl Filter for FilterStruct {
+    fn check(&mut self, _: Payload) -> bool {
+        true
+    }
+}
+
+fn main() {
+    let dims = 128;
+    let path = Path::new("/home/yanqi/stand-alone-test/data/hnsw");
```
usamoi, line 20: Where can we download files?

crates/hnsw/src/lib.rs
```diff
@@ -755,3 +756,249 @@ impl ElementHeap {
         self.binary_heap.into_iter().map(Reverse).collect()
     }
 }
+
+fn mock_make(path: &Path, options: IndexOptions) -> HnswRam<Vecf32L2> {
```
usamoi, line 760: We could use a cargo feature to decide if it's compiled.

crates/storage/src/vec.rs
```diff
@@ -13,6 +13,16 @@ pub struct VecStorage<T> {
...
+impl<T> VecStorage<T> {
+    pub fn new(vectors: MmapArray<T>, payload: MmapArray<Payload>, dims: u16) -> Self {
+        Self {
+            vectors,
+            payload,
+            dims,
+        }
+    }
+}
+
```
usamoi, line 25: It should be gated by a cargo feature too.

whateveraname: I'm not sure whether my use of cargo feature is standard and elegant, PTAL

crates/stand-alone-test/src/bin/search_hnsw.rs
whateveraname: The code currently uses the fvecs format that the sift and gist datasets use. You can refer to [http://corpus-texmex.irisa.fr/](url) for information of the fvecs format and download the sift and gist datasets. To use dataset of other formats, you may need to write another function to read that format or convert it to fvecs format.

Cargo.toml
```diff
@@ -10,11 +10,31 @@ crate-type = ["cdylib", "lib"]
...
+[[bin]]
+name = "make_hnsw"
+path = "./crates/stand-alone-test/src/bin/make_hnsw.rs"
+required-features = ["stand-alone-test"]
+
+[[bin]]
+name = "search_hnsw"
+path = "./crates/stand-alone-test/src/bin/search_hnsw.rs"
+required-features = ["stand-alone-test"]
+
```
usamoi, line 27: These binaries could be crates in package `stand-alone-test`. There is no need to modify `./Cargo.toml`.

https://doc.rust-lang.org/cargo/reference/cargo-targets.html

crates/stand-alone-test/src/bin/search_hnsw.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[derive(Clone)]
+    struct FilterStruct {}
+    impl Filter for FilterStruct {
+        fn check(&mut self, _: Payload) -> bool {
+            true
+        }
+    }
+
+    let dims = 128;
+    let path = Path::new("/home/yanqi/stand-alone-test/data/hnsw");
```
usamoi, line 26: We could use `clap` to get path of datasets from users. Otherwise anyone who wants to run these tests needs to create a user named yanqi on their machines.

crates/stand-alone-test/src/bin/make_hnsw.rs
```diff
@@ -0,0 +1,29 @@
+fn main() {
+    #[cfg(feature = "stand-alone-test")]
```
usamoi, line 2: If these binary are crates in package `stand-alone-test`, we do not need to set this cfg.

whateveraname: PTAL

PR #451, opened 2024-03-30
fix: validate_index for veci8 & rm deprecated feature.

my-vegetable-has-exploded:
I found that IndexOption check for vecint8 is lost when I merge branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,6 @@ half = { version = "2.4.0", features = [
     "num-traits",
     "rand_distr",
     "serde",
-    "use-intrinsics",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -75,7 +75,6 @@ half = { version = "2.4.0", features = [
     "num-traits",
     "rand_distr",
     "serde",
-    "use-intrinsics",
```
my-vegetable-has-exploded, line 78: This feature has been deprecated since half-2.3.0

my-vegetable-has-exploded:
thanks

Issue #459, opened 2024-04-08
feat(fdw): How to be compatible with pgvector symbol


cutecutecat:
https://github.com/pgvector/pgvector/issues/508

Possible Methods:
- Hook / Rewrite -- need verify
- Symbol Rename

cutecutecat:
# FDW problems 
At step `CREATE SERVER`:
```SQL
CREATE SERVER foreign_server
        FOREIGN DATA WRAPPER postgres_fdw
        OPTIONS (host 'host.docker.internal', port '28816', dbname 'vectors', sslmode 'disable', use_remote_estimate 'true', extensions 'vector');
GRANT USAGE ON FOREIGN SERVER foreign_server TO local_user;
```

- If **not** config the `extensions='vector'` at `CREATE SERVER` step of local machine
  - Then a full table scan will be sent to remote, unable to use an index
```SQL
postgres=> EXPLAIN verbose select * from foreign_test ORDER BY embedding <-> '[0.40671515, 0.24202824, 0.37059402]' LIMIT 1;
                                            QUERY PLAN                                             
---------------------------------------------------------------------------------------------------
 Limit  (cost=4881.00..4881.00 rows=1 width=44)
   Output: id, embedding, ((embedding <-> '[0.40671515,0.24202824,0.37059402]'::vector))
   ->  Sort  (cost=4881.00..5131.00 rows=100000 width=44)
         Output: id, embedding, ((embedding <-> '[0.40671515,0.24202824,0.37059402]'::vector))
         Sort Key: ((foreign_test.embedding <-> '[0.40671515,0.24202824,0.37059402]'::vector))
         ->  Foreign Scan on public.foreign_test  (cost=100.00..4381.00 rows=100000 width=44)
               Output: id, embedding, (embedding <-> '[0.40671515,0.24202824,0.37059402]'::vector)
               Remote SQL: SELECT id, embedding FROM public.test
```

- If config the `extensions='vector'` at `CREATE SERVER` step of local machine
  - It requires the extension at "local" and "remote" is installed in same **schema**
```SQL
postgres=> EXPLAIN verbose select * from foreign_test ORDER BY embedding <-> '[0.40671515, 0.24202824, 0.37059402]' LIMIT 1;
ERROR:  operator does not exist: vectors.vector public.<-> unknown
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
[...]

PR #466, opened 2024-04-15
feat: new text representation for sparse vector

cutecutecat:
Part of #459 

- Changed svector text representation from "[0, 1, 0, 3, 4]" to pgvector-style "{2:1, 4:3, 5:4}/5"
- Upgraded toolchain for new feature `proc_macro_byte_character` from upstream

## Reminder
The index is from ***1*** instead of 0 at pgvector

| Old                      | New                                   |
| ------------------------ | ------------------------------------- |
| [0, 1, 2, 0, 0]          | {1:1, 2:2}/5                          |
| [0, 1, 2, 3, 4, 5, 6, 7] | {1:1, 2:2, 3:3, 4:4, 5:5, 6:6, 7:7}/8 |
| [5, 6, 7]                | {0:5, 1:6, 2:7}/3                     |
| [0, 0, 0, 0]             | {}/4                                  |

### DIFF
--- a/src/datatype/text_svecf32.rs
+++ b/src/datatype/text_svecf32.rs
@@ -5,10 +5,11 @@ use base::scalar::*;
 use base::vector::*;
 use pgrx::pg_sys::Oid;
 use std::ffi::{CStr, CString};
+use std::fmt::Write;
 
 #[pgrx::pg_extern(immutable, strict, parallel_safe)]
 fn _vectors_svecf32_in(input: &CStr, _oid: Oid, _typmod: i32) -> SVecf32Output {
-    use crate::utils::parse::parse_pgvector_svector;
+    use crate::utils::parse::{parse_pgvector_svector, svector_filter_nonzero};
     let v = parse_pgvector_svector(input.to_bytes(), |s| s.parse::<F32>().ok());
     match v {
         Err(e) => {
@@ -16,7 +17,13 @@ fn _vectors_svecf32_in(input: &CStr, _oid: Oid, _typmod: i32) -> SVecf32Output {
         }
         Ok((indexes, values, dims)) => {
             check_value_dims_1048575(dims);
-            SVecf32Output::new(SVecf32Borrowed::new(dims as u32, &indexes, &values))
+            check_index_in_bound(&indexes, dims);
+            let (non_zero_indexes, non_zero_values) = svector_filter_nonzero(&indexes, &values);
+            SVecf32Output::new(SVecf32Borrowed::new(
+                dims as u32,
+                &non_zero_indexes,
+                &non_zero_values,
+            ))
         }
     }
 }
@@ -27,16 +34,16 @@ fn _vectors_svecf32_out(vector: SVecf32Input<'_>) -> CString {
     let mut buffer = String::new();
     buffer.push('{');
     let svec = vector.for_borrow();
-    let mut need_splitter = true;
+    let mut need_splitter = false;
     for (&index, &value) in svec.indexes().iter().zip(svec.values().iter()) {
         match need_splitter {
-            true => {
-                buffer.push_str(format!("{}:{}", index, value).as_str());
-                need_splitter = false;
+            false => {
+                write!(buffer, "{}:{}", index, value).unwrap();
+                need_splitter = true;
             }
-            false => buffer.push_str(format!(", {}:{}", index, value).as_str()),
+            true => write!(buffer, ", {}:{}", index, value).unwrap(),
         }
     }
-    buffer.push_str(format!("}}/{}", dims).as_str());
+    write!(buffer, "}}/{}", dims).unwrap();
     CString::new(buffer).unwrap()
 }
--- a/src/datatype/text_svecf32.rs
+++ b/src/datatype/text_svecf32.rs
@@ -1,3 +1,6 @@
+use num_traits::Zero;
+use pgrx::error;
+
 use super::memory_svecf32::SVecf32Output;
 use crate::datatype::memory_svecf32::SVecf32Input;
 use crate::error::*;
@@ -9,22 +12,66 @@ use std::fmt::Write;
 
 #[pgrx::pg_extern(immutable, strict, parallel_safe)]
 fn _vectors_svecf32_in(input: &CStr, _oid: Oid, _typmod: i32) -> SVecf32Output {
-    use crate::utils::parse::{parse_pgvector_svector, svector_filter_nonzero, svector_sorted};
+    use crate::utils::parse::parse_pgvector_svector;
     let v = parse_pgvector_svector(input.to_bytes(), |s| s.parse::<F32>().ok());
     match v {
         Err(e) => {
             bad_literal(&e.to_string());
         }
-        Ok((indexes, values, dims)) => {
-            let (mut sorted_indexes, mut sorted_values) = svector_sorted(&indexes, &values);
+        Ok((mut indexes, mut values, dims)) => {
             check_value_dims_1048575(dims);
-            check_index_in_bound(&sorted_indexes, dims);
-            svector_filter_nonzero(&mut sorted_indexes, &mut sorted_values);
-            SVecf32Output::new(SVecf32Borrowed::new(
-                dims as u32,
-                &sorted_indexes,
-                &sorted_values,
-            ))
+            // is_sorted
+            if !indexes.windows(2).all(|i| i[0] <= i[1]) {
+                assert_eq!(indexes.len(), values.len());
+                let n = indexes.len();
+                let mut permutation = (0..n).collect::<Vec<_>>();
+                permutation.sort_unstable_by_key(|&i| &indexes[i]);
+                for i in 0..n {
+                    if i == permutation[i] || usize::MAX == permutation[i] {
+                        continue;
+                    }
+                    let index = indexes[i];
+                    let value = values[i];
+                    let mut j = i;
+                    while i != permutation[j] {
+                        let next = permutation[j];
+                        indexes[j] = indexes[permutation[j]];
+                        values[j] = values[permutation[j]];
+                        permutation[j] = usize::MAX;
+                        j = next;
+                    }
+                    indexes[j] = index;
+                    values[j] = value;
+                    permutation[j] = usize::MAX;
+                }
+            }
+            let mut last: Option<u32> = None;
+            for index in indexes.clone() {
+                if last == Some(index) {
+                    error!(
+                        "Indexes need to be unique, but there are more than one same index {index}"
+                    )
+                }
+                if last >= Some(dims as u32) {
+                    error!("Index out of bounds: the dim is {dims} but the index is {index}");
+                }
+                last = Some(index);
+                {
+                    let mut i = 0;
+                    let mut j = 0;
+                    while j < values.len() {
+                        if !values[j].is_zero() {
+                            indexes[i] = indexes[j];
+                            values[i] = values[j];
+                            i += 1;
+                        }
+                        j += 1;
+                    }
+                    indexes.truncate(i);
+                    values.truncate(i);
+                }
+            }
+            SVecf32Output::new(SVecf32Borrowed::new(dims as u32, &indexes, &values))
         }
     }
 }
--- a/src/datatype/text_svecf32.rs
+++ b/src/datatype/text_svecf32.rs
@@ -1,39 +1,22 @@
 use super::memory_svecf32::SVecf32Output;
 use crate::datatype::memory_svecf32::SVecf32Input;
-use crate::datatype::typmod::Typmod;
 use crate::error::*;
 use base::scalar::*;
 use base::vector::*;
-use num_traits::Zero;
 use pgrx::pg_sys::Oid;
 use std::ffi::{CStr, CString};
 
 #[pgrx::pg_extern(immutable, strict, parallel_safe)]
-fn _vectors_svecf32_in(input: &CStr, _oid: Oid, typmod: i32) -> SVecf32Output {
+fn _vectors_svecf32_in(input: &CStr, _oid: Oid, _typmod: i32) -> SVecf32Output {
     use crate::utils::parse::parse_pgvector_svector;
-    let reserve = Typmod::parse_from_i32(typmod)
-        .unwrap()
-        .dims()
-        .map(|x| x.get())
-        .unwrap_or(0);
-    let v = parse_pgvector_svector(input.to_bytes(), reserve as usize, |s| {
-        s.parse::<F32>().ok()
-    });
+    let v = parse_pgvector_svector(input.to_bytes(), |s| s.parse::<F32>().ok());
     match v {
         Err(e) => {
             bad_literal(&e.to_string());
         }
-        Ok(vector) => {
-            check_value_dims_1048575(vector.len());
-            let mut indexes = Vec::<u32>::new();
-            let mut values = Vec::<F32>::new();
-            for (i, &x) in vector.iter().enumerate() {
-                if !x.is_zero() {
-                    indexes.push(i as u32);
-                    values.push(x);
-                }
-            }
-            SVecf32Output::new(SVecf32Borrowed::new(vector.len() as u32, &indexes, &values))
+        Ok((indexes, values, dims)) => {
+            check_value_dims_1048575(dims);
+            SVecf32Output::new(SVecf32Borrowed::new(dims as u32, &indexes, &values))
         }
     }
 }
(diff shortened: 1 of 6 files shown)


cutecutecat:
The failed CI is due to an upstream uncompatiablity: 
- proc-macro2 `v1.0.80` introduced `proc_macro_byte_character` feature, which is recently stablized
- The compiler of Github Action is built at 2024-03-26

We could wait for one of:
- Upgrade of compiler to 2024-04-xx
- The upstream fix their dependencies proc-macro2 to `v1.0.79` like

usamoi:
The reason why CI fails is that `sqllogictest-bin` released on `crates.io` but not on GitHub a few days before. We do not have to update the toolchain.

VoVAllen:
Let's hold this PR for now, due to the conflict between 1-based array and 0-based array

VoVAllen (approved):

VoVAllen:
This is used to support bm25 extension. It can produce string instead of depending on pgvecto.rs/pgvector. cc @cutecutecat

src/utils/parse.rs
```diff
@@ -83,3 +84,104 @@ where
...
+            match input[position] {
+                b'{' => break 'a position,
+                b' ' => continue,
+                _ => return Err(ParseVectorError::BadCharacter { position }),
+            }
+        }
+        return Err(ParseVectorError::BadParentheses { character: '{' });
+    };
+    let mut token: ArrayVec<u8, 48> = ArrayVec::new();
+    let mut capacity = reserve;
```
usamoi, line 112: It reserves too large since vector is sparse.

src/utils/parse.rs
```diff
@@ -83,3 +84,106 @@ where
...
+                b'}' => {
+                    token.clear();
+                    break 'a position;
+                }
+                b' ' => continue,
+                _ => return Err(ParseVectorError::BadCharacter { position }),
+            }
+        }
+        return Err(ParseVectorError::BadParentheses { character: '}' });
+    };
```
usamoi, line 136: What if `input = '{1:1,2:2,3:3}'`?

src/utils/parse.rs
```diff
@@ -83,3 +84,106 @@ where
...
+    if !token.is_empty() {
+        let position = right;
+        // Safety: all bytes in `token` are ascii characters
+        let s = unsafe { std::str::from_utf8_unchecked(&token[1..]) };
+        let num = f(s).ok_or(ParseVectorError::BadParsing { position })?;
+        indexes.push(index);
+        values.push(num);
+        token.clear();
+    }
+    Ok((indexes, values, dims))
```
usamoi, line 188: It should be sorted and removes zeros.

src/utils/parse.rs
```diff
@@ -83,3 +84,106 @@ where
...
+    }
+    if !token.is_empty() {
+        let position = right;
+        // Safety: all bytes in `token` are ascii characters
+        let s = unsafe { std::str::from_utf8_unchecked(&token[1..]) };
+        let num = f(s).ok_or(ParseVectorError::BadParsing { position })?;
+        indexes.push(index);
+        values.push(num);
+        token.clear();
+    }
```
usamoi, line 187: What if `input = '{0:}/5'`?

usamoi (changes requested):
Parsing must accept all valid inputs and reject all invalid inputs. It should be better if it can be written trivially.

src/utils/parse.rs
VoVAllen: Can we use json parsing directly?

usamoi: > Can we use json parsing directly?

Why mention json parsing?

cutecutecat: > > Can we use json parsing directly?
> 
> Why mention json parsing?

After stripped length, {1:5, 2:6, 3:7} or {} could be a valid json

usamoi: It just makes things even more complex and harder to reason correctness.

src/utils/parse.rs
cutecutecat: Added in tests.

usamoi (changes requested):
What about `{}/1/2`?

cutecutecat:
> What about `{}/1/2`?

Fixed.

usamoi:
`{1,2,3,4}/5` gives incorrect message:
```
    OutOfBound {
        dims: 5,
        index: 4294967295,
    },
```

usamoi:
`{1:1,2:2,}/5` tailing commas are not accepted, but it should be

usamoi (commented):
Since you have defined `ParseState`, you can write the code in automata style.

usamoi (changes requested):
`dims == 0` or `dims >= 2 ** 20` is not checked.
Duplicated dimension is not checked.
`value[i] == 0` is not processed.

cutecutecat:
```mermaid
graph LR
1[Start] -->|Left Bracket| 2[LeftBracket]
2[LeftBracket] -->|Number</br>backspace| 3[Index]
2[LeftBracket] -->|Right Bracket| 4[Splitter]
3[Index] -->|Number</br>push token| 3[Index]
3[Index] -->|Colon<br>collect token| 5[Value]
5[Value] -->|Number</br>push token| 5[Value]
5[Value] -->|Comma<br>collect token| 6[Comma]
5[Value] -->|Right Bracket| 4[Splitter]
6[Comma] -->|Number</br>backspace| 3[Index]
6[Comma] -->|Right Bracket| 4[Splitter]
4[Splitter] -->|"/"| 7[Length]
7[Length] -->|Number</br>push token| 7[Length]
7[Length] -->|Input end| 8(Break Loop<br>collect token)
```

usamoi:
> `dims == 0` or `dims >= 2 ** 20` is not checked. Duplicated dimension is not checked. `value[i] == 0` is not processed.

`value[i] == 0` is not processed.

src/datatype/text_svecf32.rs
```diff
@@ -1,55 +1,42 @@
...
     let mut buffer = String::new();
-    buffer.push('[');
-    let vec = vector.for_borrow().to_vec();
-    let mut iter = vec.iter();
-    if let Some(x) = iter.next() {
-        buffer.push_str(format!("{}", x).as_str());
-    }
-    for x in iter {
-        buffer.push_str(format!(", {}", x).as_str());
+    buffer.push('{');
```
usamoi, line 28: `std::fmt::Write` could be better.

src/datatype/text_svecf32.rs
```diff
@@ -1,55 +1,42 @@
...
+    let svec = vector.for_borrow();
+    let mut need_splitter = true;
```
usamoi, line 30: According to variable name, shouldn't it be `false`.

src/utils/parse.rs
```diff
@@ -83,3 +92,345 @@ where
...
+    let mut token: ArrayVec<u8, 48> = ArrayVec::new();
+
+    let mut indexes = Vec::<u32>::new();
+    let mut values = Vec::<T>::new();
+    let mut all_indexes = Vec::<u32>::new();
+    let mut index: u32 = u32::MAX;
+
+    let mut state = ParseState::Start;
+    let mut position = 0;
```
usamoi, line 127: Use `input.chars().enumerate().peekable()`

src/utils/parse.rs
```diff
@@ -83,3 +92,345 @@ where
...
+    use arrayvec::ArrayVec;
+    if input.is_empty() {
+        return Err(ParseVectorError::EmptyString {});
+    }
+    let mut token: ArrayVec<u8, 48> = ArrayVec::new();
+
+    let mut indexes = Vec::<u32>::new();
+    let mut values = Vec::<T>::new();
+    let mut all_indexes = Vec::<u32>::new();
+    let mut index: u32 = u32::MAX;
```
usamoi, line 124: Why we need this variable?

src/utils/parse.rs
```diff
@@ -83,3 +92,345 @@ where
...
+    // Check dimension out of bound
+    if dims == 0 || dims >= 1048576 {
+        return Err(ParseVectorError::InvalidDimension {
+            dims,
+            min: 0,
+            max: 1048576,
+        });
+    }
+    // Check index out of bound
+    for index in all_indexes.clone() {
```
usamoi, line 292: Why clone?

src/utils/parse.rs
```diff
@@ -83,3 +92,345 @@ where
...
+        if index as usize >= dims {
+            return Err(ParseVectorError::OutOfBound {
+                dims,
+                index: index as usize,
+            });
+        }
+    }
+    // Check index conflicts
+    let mut result: HashMap<u32, usize> = HashMap::new();
```
usamoi, line 301: `HashMap` is much slower than sort and you already performed latter.

src/utils/parse.rs
```diff
@@ -83,3 +92,345 @@ where
...
+                    b' ' => {}
+                    _ => return Err(ParseVectorError::BadCharacter { position }),
+                }
+            }
+            ParseState::LeftBracket => {
+                let c = input[position];
+                match c {
+                    b'0'..=b'9' | b'a'..=b'z' | b'A'..=b'Z' | b'.' | b'+' | b'-' => {
+                        // Do not read it here, goto Index to read
+                        position -= 1;
```
usamoi, line 148: You could just fill `token` and set next state.

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+        .map(|(_, x)| *x)
+        .collect();
+    let non_zero_values: Vec<T> = indexes
+        .iter()
+        .enumerate()
+        .filter(|(i, _)| values.get(*i).unwrap() != &T::zero())
+        .map(|(i, _)| values.get(i).unwrap().clone())
+        .collect();
+    (non_zero_indexes, non_zero_values)
+}
```
usamoi, line 117: You can just use two pointers to copy the data and call `truncate` to truncate the two arrays.

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+            position: input.len(),
+        })?;
+
+    let mut indices = (0..indexes.len()).collect::<Vec<_>>();
+    indices.sort_by_key(|&i| &indexes[i]);
+    let sorted_values: Vec<T> = indices
+        .iter()
+        .map(|i| values.get(*i).unwrap().clone())
+        .collect();
+    indexes.sort();
```
usamoi, line 237: Suggest moving it out of this function, too.

src/error.rs
```diff
@@ -69,6 +69,20 @@ ADVICE: Check if dimensions of the vector are among 1 and 1_048_575."
...
+    let mut last: u32 = 0;
+    for (i, index) in indexes.iter().enumerate() {
+        if i > 0 && last == *index {
+            error!("Indexes need to be unique, but there are more than one same index {index}")
+        }
+        if *index >= dims as u32 {
+            error!("Index out of bounds: the dim is {dims} but the index is {index}");
+        }
+        last = *index;
+    }
```
usamoi, line 82:
```suggestion
    let mut last: Option<u32> = None;
    for (i, index) in indexes.iter().enumerate() {
        if last == Some(*index) {
            error!("Indexes need to be unique, but there are more than one same index {index}")
        }
        if *index >= dims as u32 {
            error!("Index out of bounds: the dim is {dims} but the index is {index}");
        }
        last = Some(*index);
    }
```

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+            (ParseState::Start, b'{') => {
+                state = ParseState::LeftBracket;
+            }
+            (
+                ParseState::LeftBracket | ParseState::Index | ParseState::Comma,
+                b'0'..=b'9' | b'a'..=b'z' | b'A'..=b'Z' | b'.' | b'+' | b'-',
+            ) => {
+                if token.is_empty() {
+                    token.push(b'$');
+                }
```
usamoi, line 149: Why?

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+    }
+    let s = unsafe { std::str::from_utf8_unchecked(&token[1..]) };
+    let dims = s
+        .parse::<usize>()
+        .map_err(|_| ParseVectorError::BadParsing {
+            position: input.len(),
+        })?;
+
+    let mut indices = (0..indexes.len()).collect::<Vec<_>>();
+    indices.sort_by_key(|&i| &indexes[i]);
```
usamoi, line 232: I feel bad but there is surely no a function called `sort_with_indices` in std. But why call `sort` twice? A permutation can be decomposed into one or more disjoint cycles and we could use a loop to get all indexes and values in right position.

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+
+#[derive(PartialEq, Debug)]
+enum ParseState {
+    Start,
+    LeftBracket,
+    Index,
+    Value,
+    Splitter,
+    Comma,
+    Length,
```
usamoi, line 96:
```suggestion
    Dims,
```

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
 }
+
+#[derive(PartialEq, Debug)]
+enum ParseState {
+    Start,
+    LeftBracket,
+    Index,
+    Value,
+    Splitter,
+    Comma,
```
usamoi, line 95:
```suggestion
    Comma,
    Splitter,
```

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+                if token.is_empty() {
+                    return Err(ParseVectorError::TooShortNumber { position });
+                }
+                let s = unsafe { std::str::from_utf8_unchecked(&token[1..]) };
+                let index = s
+                    .parse::<u32>()
+                    .map_err(|_| ParseVectorError::BadParsing { position })?;
+                indexes.push(index);
+                token.clear();
+                state = ParseState::Value;
```
usamoi, line 168: There should be a state, if the last input character is `:`. And we do not need to check `token.is_empty()` because if parser does not see digitals it does not enter `Index`, `Value` or `Dims`.

src/utils/parse.rs
```diff
@@ -83,3 +84,286 @@ where
...
+        return Err(ParseVectorError::EmptyString {});
+    }
+    let mut token: ArrayVec<u8, 48> = ArrayVec::new();
+    let mut indexes = Vec::<u32>::new();
+    let mut values = Vec::<T>::new();
+
+    let mut state = ParseState::Start;
+    for (position, char) in input.iter().enumerate() {
+        let c = *char;
+        match (&state, c) {
```
usamoi, line 138:
```suggestion
    for (position, c) in input.iter().copied().enumerate() {
        state = match (state, c) {
```

usamoi (commented):
It should be correct.

VoVAllen:
Please get this merge today, thanks

src/datatype/text_svecf32.rs
```diff
@@ -1,55 +1,50 @@
...
-                    indexes.push(i as u32);
-                    values.push(x);
-                }
-            }
-            SVecf32Output::new(SVecf32Borrowed::new(vector.len() as u32, &indexes, &values))
+        Ok((indexes, values, dims)) => {
+            let (mut sorted_indexes, mut sorted_values) = svector_sorted(&indexes, &values);
+            check_value_dims_1048575(dims);
+            check_index_in_bound(&sorted_indexes, dims);
+            svector_filter_nonzero(&mut sorted_indexes, &mut sorted_values);
```
usamoi, line 22:
```suggestion
            check_value_dims_1048575(dims);
            if !indexes.is_sorted() {
                assert_eq!(indexes.len(), values.len());
                let n = indexes.len();
                let mut permutation = (0..n).collect::<Vec<_>>();
                permutation.sort_unstable_by_key(|&i| &indexes[i]);
                for i in 0..n {
                    if i == permutation[i] || usize::MAX == permutation[i] {
                        continue;
                    }
                    let index = indexes[i];
                    let value = values[i];
                    let mut j = i;
                    while i != permutation[j] {
                        let next = permutation[j];
                        indexes[j] = indexes[permutation[j]];
                        values[j] = values[permutation[j]];
                        permutation[j] = usize::MAX;
                        j = next;
                    }
                    indexes[j] = index;
                    values[j] = value;
                    permutation[j] = usize::MAX;
                }
            }
            let mut last = None;
            for index in indexes {
                if last == Some(*index) {
                    error!("Indexes need to be unique, but there are more than one same index {index}")
            }
            if last >= Some(dims) {
                error!("Index out of bounds: the dim is {dims} but the index is {index}");
            }
            {
                let mut i = 0;
                let mut j = 0;
                while j < values.len() {
                    if !values[j].is_zero() {
                        indexes[i] = indexes[j];
                        values[i] = values[j].clone();
                        i += 1;
                    }
                    j += 1;
                }
                indexes.truncate(i);
                values.truncate(i);
            }
```
And remove these three functions.

usamoi (approved):

PR #469, opened 2024-04-19
fix: fix openai option gucs setting.

my-vegetable-has-exploded:
I think those code don't matchs.
```rust     
let base_url = guc_string_parse(&OPENAI_BASE_URL, "vectors.openai_base");
```
and 

```rust
    GucRegistry::define_string_guc(
        "vectors.openai_base_url",
        "The base url of OpenAI or compatible server.",
        "",
        &OPENAI_BASE_URL,
        GucContext::Userset,
        GucFlags::default(),
    );
```

### DIFF
--- a/src/gucs/embedding.rs
+++ b/src/gucs/embedding.rs
@@ -4,7 +4,7 @@ use pgrx::{GucContext, GucFlags, GucRegistry, GucSetting};
 use std::ffi::CStr;
 
 pub fn openai_options() -> OpenAIOptions {
-    let base_url = guc_string_parse(&OPENAI_BASE_URL, "vectors.openai_base");
+    let base_url = guc_string_parse(&OPENAI_BASE_URL, "vectors.openai_base_url");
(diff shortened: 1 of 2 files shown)


src/gucs/embedding.rs
```diff
@@ -13,7 +13,7 @@ static OPENAI_API_KEY: GucSetting<Option<&'static CStr>> =
     GucSetting::<Option<&'static CStr>>::new(None);
 
 static OPENAI_BASE_URL: GucSetting<Option<&'static CStr>> =
-    GucSetting::<Option<&'static CStr>>::new(Some(c"https://api.openai.com/v1/"));
+    GucSetting::<Option<&'static CStr>>::new(Some(c"https://api.openai.com/v1"));
```
my-vegetable-has-exploded, line 16: And this don't match our code in `embedding/src/lib.rs`. 
```    let url = format!("{}/embeddings", opt.base_url);```

cutecutecat (approved):

Issue #475, opened 2024-05-10
Feature Request: Add Sum Aggregation and Column-Wise Multiplication for Sparse Vectors


gioeleperin:
Starting off, we'd like to express our appreciacion for the project. 

Integrating the following operations for sparse vectors would be of great use to our use case:
- An aggregate sum operation for sparse vectors, similar to the one already implemented for dense vectors.
- Element-wise multiplication.

By incorporating these functionalities, we can streamline computations directly within the database, sparing us the overhead of loading them into our application. 

Thank you for considering this request.

VoVAllen:
Thanks for your interest! May I ask what's your scenario to do the aggregate sum for sparse vectors? What's the query looks like?

VoVAllen:
And what's  Column-Wise Multiplication? Do you mean dot product between sparse vectors?

VoVAllen:
Understand! PRs on the way

PR #463, opened 2024-04-09
feat: support aggregate function for vector.

my-vegetable-has-exploded:
close #458 

This pr mainly contains function including `avg` ，`sum`, `vector_dims`, `vector_norm`.

### vector_dims(vector)

Returns the dimensions of a given vector.

Arguments：a vector.
Return type integer representing the number of dimensions of the given vector.

### vector_norms(vector)

Calculates the Euclidean norm of a given vector.
Arguments:a vector.
Return double precision representing the Euclidean norm of the given vector.

### AVG

Calculates the average of the processed vectors.
Arguments
vector:  vectors.
Return vector representing the average of processed vectors.

### SUM
Arguments： vectors.
Return type vector representing the sum of processed vectors.

The aggregate function support need to add a state transition function sfunc and an optional final calculation function ffunc, following [create aggregate document](https://www.postgresql.org/docs/current/sql-createaggregate.html).  And we add `COMBINEFUNC`  here to support parallel aggregate.

PR #476, opened 2024-05-12
Implement aggregate functions for sparse vector.

my-vegetable-has-exploded:
refs #475

Purpose of this pr is almost same with #463, but mainly for sparse vector.


src/sql/finalize.sql
```diff
@@ -86,6 +86,11 @@ CREATE TYPE vector_accumulate_state (
     ALIGNMENT = double
 );
 
+CREATE TYPE svector_accumulate_state AS (
```
my-vegetable-has-exploded, line 89: I just use composite type rather than a custom type with in/out function here. It is not easy to reuse the state for sparse vector (length of `index` and `value` may change), so I think it is just okay to implement avg using plpgsql.

src/sql/finalize.sql
```diff
@@ -625,6 +681,22 @@ CREATE AGGREGATE sum(vector) (
     PARALLEL = SAFE
 );
 
+CREATE AGGREGATE avg(svector) (
+	SFUNC = _vectors_svector_accum,
+	STYPE = svector_accumulate_state,
+	COMBINEFUNC = _vectors_svector_combine,
+	FINALFUNC = _vectors_svector_final,
+	INITCOND = '(0, [0])',
```
my-vegetable-has-exploded, line 689: I tried to use `(0, NULL`). But I don't find a way to use `NULL` for `INITCOND`;

VoVAllen:
@usamoi PTAL

src/sql/finalize.sql
usamoi: We could just use an array length of which is `dims` since the result is probably dense.

src/sql/finalize.sql
usamoi: `(0,)`

my-vegetable-has-exploded: get! thanks.

src/sql/finalize.sql
my-vegetable-has-exploded: I am not sure about it. Do you think it is reasonable？ @gioeleperin

gioeleperin: > I am not sure about it. Do you think it is reasonable？ @gioeleperin

I'm not sure, as I don't have a deep understanding of how the underlying system works. However, I can say that when performing aggregations on sparse vectors, such as summing or averaging, the resulting vector is expected to remain sparse.

usamoi: If the the aggregation on sparse vectors is a sparse vector, we can perform a strategy that doubles the length of buffer if needed.

VoVAllen: it's still sparse. Can we specify buffer length here manually?

usamoi: > it's still sparse. Can we specify buffer length here manually?

The main problem is that we do know the length of the aggregation so we cannot allocate a buffer of proper size.

my-vegetable-has-exploded: > If the the aggregation on sparse vectors is a sparse vector, we can perform a strategy that we double the length of buffer if needed so that we can reuse the state for most iterations.

good idea.

VoVAllen:
Is this ready to merge?

my-vegetable-has-exploded:
> Is this ready to merge?

not yet. I am rewriting the state amd related function.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,454 @@
...
+                .write(SparseAccumulateStateHeader::varlena(layout.size()));
+            std::ptr::addr_of_mut!((*ptr).dims).write(dims);
+            std::ptr::addr_of_mut!((*ptr).len).write(0);
+            std::ptr::addr_of_mut!((*ptr).capacity).write(capacity as u32);
+            std::ptr::addr_of_mut!((*ptr).count).write(0);
+            SparseAccumulateState::Owned(NonNull::new(ptr).unwrap())
+        }
+    }
+
+    pub fn merge_in_place(&mut self, svec: SVecf32Borrowed<'_>) -> Result<(), AccumulateError> {
```
my-vegetable-has-exploded, line 133: This main difference from last commit. This function try to merge a sparse vector into its state in-place.

I'd like to write more unit tests for it. But it don't works to add `mod tests` just in this file. Could you give me some hints?@usamoi

usamoi: It should work if you add `mod tests` gated with `#[cfg(test)]` because the linker will eliminate dead code so that `test` binary is not linked with PostgreSQL. If you worry about producing a `SparseAccumulateState`, you can create a new type without any PostgreSQL stuff.

VoVAllen:
@usamoi PTAL

src/sql/finalize.sql
```diff
@@ -607,24 +615,50 @@ STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_vector_dims_wrap
...
+    FINALFUNC = _vectors_vector_final_avg,
     INITCOND = '0, []',
     PARALLEL = SAFE
 );
 
 CREATE AGGREGATE sum(vector) (
-    SFUNC = _vectors_vecf32_operator_add,
-    STYPE = vector,
-    COMBINEFUNC = _vectors_vecf32_operator_add,
+	SFUNC = _vectors_vector_accum,
```
my-vegetable-has-exploded, line 636: Sorry, I will handle those `tap` error later.

src/datatype/functions_svecf32.rs
```diff
@@ -43,3 +540,163 @@ fn _vectors_to_svector(
     }
     SVecf32Output::new(SVecf32Borrowed::new(dims.get(), &indexes, &values))
 }
+
+/// divide a sparse vector by a scalar.
+#[pgrx::pg_extern(immutable, strict, parallel_safe)]
+fn _vectors_svecf32_div(vector: SVecf32Input<'_>, scalar: f32) -> SVecf32Output {
```
usamoi, line 546: 1. where is it used?
2. if scalar is very large, and values in vector are small, `values` contains zeros so `SVecf32Output::new` panics. You can add a function to receive input containing zeros as a helper.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,504 @@
...
+#[pgrx::pg_extern(immutable, strict, parallel_safe)]
+fn _vectors_svector_final_sum(state: SparseAccumulateState<'_>) -> Option<SVecf32Output> {
+    let count = state.count();
+    if count == 0 {
+        // return NULL if all inputs are NULL
+        None
+    } else {
+        let indexes = state.indexes();
+        let values = state.values();
+        Some(SVecf32Output::new(SVecf32Borrowed::new(
```
usamoi, line 495: Same problem here: it may panic.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,504 @@
...
+        // return NULL if all inputs are NULL
+        None
+    } else {
+        let len = state.len();
+        state.indexes_values_mut().1[..len]
+            .iter_mut()
+            .for_each(|x| *x /= count as f32);
+        let indexes = state.indexes();
+        let values = state.values();
+        Some(SVecf32Output::new(SVecf32Borrowed::new(
```
usamoi, line 477: Same problem here.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,504 @@
...
+        if let Some(x) = iter.next() {
+            buffer.push_str(format!("{}", x).as_str());
+        }
+        for x in iter {
+            buffer.push_str(format!(", {}", x).as_str());
+        }
+    }
+    buffer.push(']');
+    CString::new(buffer).unwrap()
+}
```
usamoi, line 377: It could be simpler since it's internal and no user would use it.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,504 @@
...
+            capacity,
+            count,
+            indexes,
+            values,
+        }
+    }
+
+    /// Try to merge a sparse vector into the state in place.
+    /// Return the length of the merged state if success, otherwise return an error.
+    pub fn merge_in_place(&mut self, svec: SVecf32Borrowed<'_>) -> Result<u32, AccumulateError> {
```
usamoi, line 114: We could grow the state util there are more empty slows than values in `svec` so that `merge_in_place` never fails. It's simpler.

src/datatype/functions_svecf32.rs
```diff
@@ -1,7 +1,504 @@
...
+                .copy_within(((p + 1) as usize)..capacity, (i as usize) + 1);
+            self.values
+                .copy_within(((p + 1) as usize)..capacity, (i as usize) + 1);
+            ((capacity as i32) - p + i) as u32
+        };
+        Ok(self.len)
+    }
+}
+
+fn next_power_of_two(a: usize) -> usize {
```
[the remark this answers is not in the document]
my-vegetable-has-exploded: good idea.

src/datatype/functions_svecf32.rs
my-vegetable-has-exploded: Right, it makes the code more readable.

src/datatype/functions_svecf32.rs
my-vegetable-has-exploded: > 1. where is it used?

It was used in old implementation. It is useless now. 

>     2. if scalar is very large, and values in vector are small, `values` contains zeros so `SVecf32Output::new` panics. You can add a function to receive input containing zeros as a helper.

Good catch! Thanks.

my-vegetable-has-exploded:
ptal @usamoi

usamoi (dismissed):
It looks fine but too complex to review.

VoVAllen:
Since it's hard to review, can @my-vegetable-has-exploded add more randomized tests and run multiple times locally to ensure the correctness?

VoVAllen:
Both end2end and unit test

my-vegetable-has-exploded:
> Since it's hard to review, can @my-vegetable-has-exploded add more randomized tests and run multiple times locally to ensure the correctness?

Okay, I'll also look into how to make the code more readable.

tests/sqllogictest/svector.slt
```diff
@@ -0,0 +1,165 @@
...
+[1, 1, 1, 1, 1]
+
+query ?
+SELECT avg(v) FROM unnest(ARRAY[to_svector(32, '{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15}', '{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}'), to_svector(32, '{1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16}', '{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}'), to_svector(32, '{2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17}', '{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}') ]) v;
+----
+[0.33333334, 0.6666667, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0.6666667, 0.33333334, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+
+# test avg(svector) get the same result as avg(vector)
+query ?
+SELECT avg(data) = avg(data::svector)::vector FROM test_vectors;
```
my-vegetable-has-exploded, line 109: The elements from avg(data) is essentially within the range of [2,3],  matching the expected value of the generated data ((100+1)/2*0.05) ~= 2.5.

src/datatype/aggregate_svecf32.rs
```diff
@@ -0,0 +1,792 @@
...
+        };
+        #[allow(unused_unsafe)]
+        let result = unsafe { _vectors_svecf32_aggregate_avg_sum_combinefunc(state1_, state2_) };
+        ::pgrx::datum::IntoDatum::into_datum(result)
+            .unwrap_or_else(|| panic!("returned Datum was NULL"))
+    }
+    #[allow(unused_unsafe)]
+    unsafe {
+        pgrx::pg_sys::submodules::panic::pgrx_extern_c_guard(move || {
+            let mut agg_context: *mut ::pgrx::pg_sys::MemoryContextData = std::ptr::null_mut();
```
my-vegetable-has-exploded, line 614: pg will switch to a temporary memory context when call aggregate trans function.
https://github.com/postgres/postgres/blob/52b49b796cc7fd976f4da6aa49c9679ecdae8bd5/src/backend/executor/nodeAgg.c#L761-L801
[doxygen](https://doxygen.postgresql.org/nodeAgg_8c.html#a8756447ac100aaea4999db35c4280939)
So we need to switch context like https://github.com/postgres/postgres/blob/7c655a04a2dc84b59ed6dce97bd38b79e734ecca/src/backend/utils/adt/numeric.c#L5635-L5648.
Currently pgrx's `Aggregate` trait can't pass a reference with lifetime. https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx/src/aggregate.rs. So I expand it. （ Of course, this piece of code is terrible.）

my-vegetable-has-exploded: https://github.com/pgcentralfoundation/pgrx/issues/1719

usamoi: I feel not necessary for using `Aggregate` trait.

my-vegetable-has-exploded: I don't find a good way to swtich the context under pgrx currently.😭

usamoi: You could just write code in C way. `pgrx` is more a binding than a complete interface of PostgreSQL.

my-vegetable-has-exploded: > You could just write code in C way. `pgrx` is more a binding than a complete interface of PostgreSQL.

I add a macro to wrap it currently.

src/datatype/aggregate_svecf32.rs
```diff
@@ -0,0 +1,485 @@
...
+
+/// accumulate intermediate state for sparse vector
+#[base_macros::aggregate_func]
+#[pgrx::pg_extern(immutable, parallel_safe)]
+fn _vectors_svecf32_aggregate_avg_sum_sfunc(
+    mut current: Option<Internal>,
+    value: Option<SVecf32Input<'_>>,
+) -> Option<Internal> {
+    if value.is_none() {
+        // It would get error "returned Datum was NULL" if we use Internal directly when execute `SELECT sum(v) FROM unnest(ARRAY[NULL]::svector[]) v;`
```
my-vegetable-has-exploded, line 198: It seems upstream have solve it. https://github.com/pgcentralfoundation/pgrx/pull/1701

usamoi: You could submit a PR to use branch `v0.12.0-alpha.1-patch2`.

src/sql/finalize.sql
```diff
@@ -664,83 +651,46 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_binari
...
-	IF count = 0 THEN
-		RETURN NULL;
-	END IF;	
-	result := _vectors_svecf32_div(state.sum, count::real);
-	RETURN result;
-END;
-$$ LANGUAGE plpgsql STRICT PARALLEL SAFE;
+CREATE FUNCTION _vectors_svecf32_aggregate_avg_sum_sfunc_agg(internal, svector) RETURNS internal IMMUTABLE PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_svecf32_aggregate_avg_sum_sfunc_agg_wrapper';
+
+CREATE FUNCTION _vectors_svecf32_aggregate_avg_sum_combinefunc_agg(internal, internal) RETURNS internal IMMUTABLE PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_svecf32_aggregate_avg_sum_combinefunc_agg_wrapper';
```
usamoi, line 660: You could use `pgrx::pg_extern(sql = "...")` to embed it.

src/datatype/mod.rs
```diff
@@ -32,3 +33,11 @@ pub mod text_vecf16;
 pub mod text_vecf32;
 pub mod text_veci8;
 pub mod typmod;
+
+use pgrx::Internal;
+
+fn get_mut_internal<T>(internal: &mut Option<Internal>) -> Option<&mut T> {
```
usamoi, line 39: It cannot be marked safe.

crates/base_macros/src/lib.rs
```diff
@@ -75,3 +75,53 @@ pub fn alter(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    pub extern "C" fn #pg_finfo_fn(
+    ) -> &'static ::pgrx::pg_sys::Pg_finfo_record {
+        const V1_API: ::pgrx::pg_sys::Pg_finfo_record =
+            ::pgrx::pg_sys::Pg_finfo_record { api_version: 1 };
+        &V1_API
+    }
+    };
+
+    proc_macro::TokenStream::from(expanded)
+}
```
usamoi, line 127: Why must we use such a macro?

src/sql/finalize.sql
my-vegetable-has-exploded: get

src/datatype/mod.rs
my-vegetable-has-exploded: May I mark this function unsafe? Inline it would make code below not so readable?
```rust
fn _vectors_svecf32_aggregate_avg_sum_combinefunc(
    mut state1: Option<Internal>,
    mut state2: Option<Internal>,
) -> Option<Internal> {
    match (
        get_mut_internal::<SVecf32AggregateAvgSumStype>(&mut state1),
        get_mut_internal::<SVecf32AggregateAvgSumStype>(&mut state2),
    ) {
```

my-vegetable-has-exploded: > May I mark this function unsafe? Inline it would make this `match` statement to long.
> 
> ```rust
> fn _vectors_svecf32_aggregate_avg_sum_combinefunc(
>     mut state1: Option<Internal>,
>     mut state2: Option<Internal>,
> ) -> Option<Internal> {
>     match (
>         get_mut_internal::<SVecf32AggregateAvgSumStype>(&mut state1),
>         get_mut_internal::<SVecf32AggregateAvgSumStype>(&mut state2),
>     ) {
> ```

crates/base_macros/src/lib.rs
my-vegetable-has-exploded: I think it is easier than writing memory context switch codes for every aggregate function.

usamoi: Is there any difference if you use `pg_extern` and switch contexts manually?

src/datatype/mod.rs
usamoi: > Inline it would make this match statement to long.

You can use `let ...` first.

crates/base_macros/src/lib.rs
my-vegetable-has-exploded: `AggCheckCallContext` need use `fcinfo`,  so we can only switch in wrapper function. I don't know how to switch contexts in our rust function.

usamoi: You could get fcinfo by adding a parameter `fcinfo: pgrx::pg_sys::FunctionCallInfo`. It's an undocumented behavior in pgrx.

my-vegetable-has-exploded: great！ Really a batter way.

my-vegetable-has-exploded: great！ Really a batter way.

my-vegetable-has-exploded:
ptal @usamoi

src/datatype/aggregate_svecf32.rs
```diff
@@ -0,0 +1,492 @@
...
+            // concatenate the state [p+1..capacity] and remaining state [0..i]
+            self.indexes
+                .copy_within(((p + 1) as usize)..capacity, (i as usize) + 1);
+            self.values
+                .copy_within(((p + 1) as usize)..capacity, (i as usize) + 1);
+            ((capacity as i32) - p + i) as u32
+        };
+        self.len
+    }
+}
```
usamoi, line 187: Why we need this type?

my-vegetable-has-exploded: It was used for ut in older implementation. I will remove it.

src/datatype/aggregate_svecf32.rs
```diff
@@ -0,0 +1,428 @@
...
+                // allocate a new state and merge the old state
+                let mut new_state = SVecf32AggregateAvgSumStype::new_with_capacity(
+                    dims as u32,
+                    state.len() + value.len(),
+                );
+                new_state.merge_in_place(SVecf32Borrowed::new(
+                    dims as u32,
+                    state.indexes(),
+                    state.values(),
+                ));
```
usamoi, line 192: `state` may not be a `SVecf32` because of zeros. `merge_in_place` should not accept `SVecf32Borrowed`.

my-vegetable-has-exploded: I think zero number will be filter in Line115.

usamoi: So what `filter_zero` is used for?

my-vegetable-has-exploded: Mainly for `avg` aggregate. Elements in `state` may be get zero after division by `count`.

src/datatype/aggregate_svecf32.rs
```diff
@@ -0,0 +1,428 @@
...
+        None => None,
+    }
+}
+
+/// finalize the intermediate state for sparse vector sum
+#[pgrx::pg_extern(immutable, strict, parallel_safe)]
+fn _vectors_svecf32_aggregate_sum_finalfunc(state: Internal) -> Option<SVecf32Output> {
+    match unsafe { state.get_mut::<SVecf32AggregateAvgSumStype>() } {
+        Some(state) => {
+            state.filter_zero();
```
usamoi, line 304: So it can be safely removed.

my-vegetable-has-exploded: It should be reasonable.

usamoi (approved):

PR #481, opened 2024-05-20
fix: use f*_algebraic intrinsics, instead of f*_fast intrinsics

usamoi:
f*_algebraic intrinsics are safer than f*_fast intrinsics since they do not assume inputs are not NaNs or Inf.


usamoi:
it generates same asm: https://godbolt.org/z/bndTTGrEb

PR #489, opened 2024-05-28
feat: temporary support for sparse aggregate. 

my-vegetable-has-exploded:
Temporary support for sparse aggregate.  This will be replace by #476.


VoVAllen:
@usamoi PTAL. Let's merge a simple one first.

usamoi (approved):

PR #505, opened 2024-06-19
chore: remove prefilter, stand-alone-test and default feature

usamoi:
This pull request
* removes prefilter: the implementation is not usable at all and blocks future work
* removes `stand-alone-test`: it's just a copy of `hnsw`
* removes default feature `pg15` in `Cargo.toml` to avoid too many `--no-default-feature`s for whom does not install PostgreSQL 15.

If you're developing with PostgreSQL 16, after this pr is merged, you need to add following json snippet to `.vscode/settings.json`.

```
"rust-analyzer.cargo.features": [
    "pg16"
],
```


usamoi:
It does not add any new code, can it be merged (without waiting before `0.3.0` releases)? @VoVAllen

Issue #509, opened 2024-06-22
feat: flat sparse vector index using column oriented layout


VoVAllen:
https://github.com/xhluca/bm25s showed great performance to use brute force approache for bm25 indexing. We can adopt it into pgvecto.rs

PR #517, opened 2024-07-04
feat: support inverted index for sparse vector

kemingy:
- close https://github.com/tensorchord/pgvecto.rs/issues/509
- TODO: https://github.com/tensorchord/pg_bestmatch.rs/issues/16


crates/base/src/vector/svecf32.rs
```diff
@@ -171,6 +171,11 @@ impl<'a> VectorBorrowed for SVecf32Borrowed<'a> {
         }
     }
 
+    #[inline(always)]
+    fn to_index_vec(&self) -> Vec<(u32, Self::Scalar)> {
+        std::iter::zip(self.indexes.to_vec(), self.values.to_vec()).collect()
+    }
```
usamoi, line 177: This function should be moved to `OperatorInverted` in `./crates/inverted/operator.rs`. The implementation for all other types should be `unimplemented!()`.

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+impl<O: OperatorInverted> Inverted<O> {
+    pub fn create(path: impl AsRef<Path>, options: IndexOptions, source: &impl Source<O>) -> Self {
+        if options.vector.v != VectorKind::SVecf32 {
+            panic!("inverted index only supports `SVecf32` vectors");
+        }
+        if options.vector.d != DistanceKind::Dot {
+            panic!("inverted index only supports `Dot` distance");
+        }
```
usamoi, line 42: It should be checked in `validate`. Putting checks here will make building index fail silently.

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+use std::path::Path;
+
+pub trait OperatorInverted: Operator + OperatorQuantization + OperatorStorage {}
+
+impl<T: Operator + OperatorQuantization + OperatorStorage> OperatorInverted for T {}
+
+#[allow(dead_code)]
+pub struct Inverted<O: OperatorInverted> {
+    storage: O::Storage,
+    quantization: Quantization<O>,
```
usamoi, line 28:
```suggestion
```

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        vector: Borrowed<'a, O>,
+        opts: &'a SearchOptions,
+    ) -> (Vec<Element>, Box<(dyn Iterator<Item = Element> + 'a)>) {
+        const ZERO: F32 = F32(0.0);
+        let mut doc_score = vec![ZERO; self.payloads.len()];
+        for (token, _) in vector.to_index_vec() {
+            let start = self.offsets[token as usize];
+            let end = self.offsets[token as usize + 1];
+            for i in (start as usize)..(end as usize) {
+                doc_score[self.indexes[i] as usize] += self.scores[i];
```
usamoi, line 62: Why it's not `doc_score[self.indexes[i] as usize] += self.scores[i] * val`?

kemingy: `val` is not used here since the sparse matrix already stores the BM25 score. Only the indices are used.

usamoi: Do you mean `vector` is a binary vector?

If it's not guaranteed, and when there are two segments (one is indexed and one is not indexed), the index returns incorrect results because bruteforce in growing segments performs `* val`.

crates/inverted/src/lib.rs
usamoi:
```rust
let mut documents = (0..self.storage.len()).map(|i| Element { distance: F32::zero(), payload: self.payload[i as usize] }).collect::<Vec<_>>();
// ...
documents.sort();
(Vec::new(), Box::new(documents.into_iter())
```

crates/index/Cargo.toml
```diff
@@ -28,6 +28,7 @@ storage = { path = "../storage" }
 # algorithms
 flat = { path = "../flat" }
 hnsw = { path = "../hnsw" }
+inverted = { path = "../inverted" }
```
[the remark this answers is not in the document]
usamoi: No. @VoVAllen Do you have a good name?

crates/inverted/src/lib.rs
kemingy: I'm ok with that if `reranker` is not compulsory in the implementation (which makes it different from others).

crates/inverted/src/lib.rs
usamoi: How about checking if `val` is 1.0 for first? If it is used correctly, the slow path is never accessed. If it is used incorrectly, it still returns correct results.

crates/base/src/vector/svecf32.rs
kemingy: fixed

crates/inverted/src/lib.rs
kemingy: I think `pg_bestmatch` can change the doc score to bm25 and the query score to 0-1. Both dot-production and inverted-index work in the same way.

crates/inverted/src/operator.rs
```diff
@@ -0,0 +1,37 @@
...
+use base::scalar::F32;
+use quantization::operator::OperatorQuantization;
+use storage::OperatorStorage;
+
+pub trait OperatorInverted: OperatorQuantization + OperatorStorage {
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> Vec<(u32, F32)>;
+}
+
+impl OperatorInverted for SVecf32Dot {
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> Vec<(u32, F32)> {
```
usamoi, line 11:
```suggestion
    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> {
```

crates/base/src/index.rs
```diff
@@ -259,6 +261,7 @@ pub enum IndexingOptions {
     Flat(FlatIndexingOptions),
     Ivf(IvfIndexingOptions),
     Hnsw(HnswIndexingOptions),
+    Inverted(),
```
usamoi, line 264: There should be `InvertedIndexingOptions` even if it's empty.

usamoi (approved):
Names like `inverted_file_bm25` or `inverted_file_sparse` may be better.

crates/inverted/src/operator.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use std::iter::{zip, Empty};
+
+pub trait OperatorInvertedSparse: OperatorQuantization + OperatorStorage {
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)>;
+}
+
+impl OperatorInvertedSparse for SVecf32Dot {
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> {
+        zip(vec.indexes().to_vec(), vec.values().to_vec())
```
usamoi, line 14: `std::iter::zip(vec.indexes().iter(), vec.values().iter())`

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let storage = O::Storage::create(path.as_ref().join("storage"), collection);
+    let payloads = MmapArray::create(
+        path.as_ref().join("payloads"),
+        (0..collection.len()).map(|i| collection.payload(i)),
+    );
+    let json_index = Json::create(path.as_ref().join("indexes"), indexes);
+    let json_offset = Json::create(path.as_ref().join("offsets"), offsets);
+    let json_score = Json::create(path.as_ref().join("scores"), scores);
+    sync_dir(path);
```
usamoi, line 112: No need to call it manually because of a recent commit.

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+fn from_nothing<O: OperatorInvertedSparse>(
+    path: impl AsRef<Path>,
+    _: IndexOptions,
+    collection: &impl Collection<O>,
+) -> InvertedSparse<O> {
+    create_dir(path.as_ref()).expect("failed to create path for inverted sparse index");
+
+    let mut token_collection = BTreeMap::new();
```
usamoi, line 93:
```suggestion
    let mut token_collection = Vec::new();
```

crates/inverted/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+            let start = self.offsets[token as usize];
+            let end = self.offsets[token as usize + 1];
+            for i in (start as usize)..(end as usize) {
+                doc_score[self.indexes[i] as usize] += self.scores[i] * val;
+            }
+        }
+        let mut candidates: BinaryHeap<(F32, Payload)> = doc_score
+            .iter()
+            .enumerate()
+            .filter(|&(_, score)| *score > ZERO)
```
usamoi, line 57: Why?

usamoi:
`InvertedSparse`. `Sparse` is even not a noun...

crates/inverted/src/operator.rs
usamoi:
```
std::iter::zip(vec.indexes().iter().copied(), vec.values().iter().copied())
```

crates/inverted/src/lib.rs
usamoi: No matter if it's meaningful, it should be returned.

crates/inverted/src/lib.rs
usamoi:
```
let mut token_collection = vec![Vec::new(); dims as usize];
```

crates/inverted/src/operator.rs
kemingy: borrowed value does not live long enough

usamoi:
```
fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> + '_;
```

crates/inverted/src/lib.rs
kemingy: Why should we return the documents that are not relevant at all?

usamoi: A query should return all tuples in order if there are only `ORDER BY` clauses according to semantics of SQL.

crates/inverted/src/operator.rs
kemingy: `vec` is borrowed, returns a value referencing data owned by the current function

usamoi:
```
diff --git a/crates/base/src/vector/svecf32.rs b/crates/base/src/vector/svecf32.rs
index be71b5f..6a2934f 100644
--- a/crates/base/src/vector/svecf32.rs
+++ b/crates/base/src/vector/svecf32.rs
@@ -138,12 +138,12 @@ impl<'a> SVecf32Borrowed<'a> {
     }
 
     #[inline(always)]
-    pub fn indexes(&self) -> &[u32] {
+    pub fn indexes(&self) -> &'a [u32] {
         self.indexes
     }
 
     #[inline(always)]
-    pub fn values(&self) -> &[F32] {
+    pub fn values(&self) -> &'a [F32] {
         self.values
     }
 
diff --git a/crates/inverted/src/operator.rs b/crates/inverted/src/operator.rs
index 56af6d7..6dd414a 100644
--- a/crates/inverted/src/operator.rs
+++ b/crates/inverted/src/operator.rs
@@ -6,12 +6,12 @@ use storage::OperatorStorage;
 use std::iter::{zip, Empty};
 
 pub trait OperatorInvertedIndex: OperatorQuantization + OperatorStorage {
-    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)>;
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> + '_;
 }
 
 impl OperatorInvertedIndex for SVecf32Dot {
-    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> {
-        zip(vec.indexes().to_vec(), vec.values().to_vec())
+    fn to_index_vec(vec: Borrowed<'_, Self>) -> impl Iterator<Item = (u32, F32)> + '_ {
+        zip(vec.indexes().iter().copied(), vec.values().iter().copied())
     }
 }
```

kemingy: Fixed

usamoi (approved):

PR #525, opened 2024-07-11
chore: replace uuid with integer for segment id

usamoi:


VoVAllen:
Did you bump the format version?

PR #533, opened 2024-07-19
feat: kmeans1d

usamoi:


VoVAllen:
Why do we need kmeans1d?

usamoi:
> Why do we need kmeans1d?

It can be used in product quantization.

PR #534, opened 2024-07-22
feat: quantization, 1/2/4 bits

usamoi:
blocked by #533


VoVAllen:
Did you figure out the precision problem?

usamoi:
> Did you figure out the precision problem?

No. But it's not related to 1/2/4 bits.

PR #537, opened 2024-07-23
fix: quantization precision

usamoi:
It's still a little worser than faiss-ivfpq.

**This pr deprecates veci8 by the way.**


VoVAllen:
What's the precision problem?

usamoi:
> What's the precision problem?

Spherical k-means is used for both product quantization and IVF before. But it should not be used in product quantization. This pr removes normalization in k-means.

VoVAllen (approved):

PR #555, opened 2024-08-12
refactor: remove indexing on cosine similarity on backend

usamoi:
Implementing indexing on cosine similarity by normalizing indexing vectors and query vectors. It helps fast scan implementation and speeds computing cosine similarity.


VoVAllen:
please add some tests

PR #557, opened 2024-08-13
test: fix cos test by comparing shorter result list

usamoi:
`hnsw_ef_search` is 100, so we cannot get exact top-100 KNN search results.


VoVAllen:
Please add description in PR about why LIMIT 100 may cause error

PR #559, opened 2024-08-14
fix: add v4, v3 for vecf32 dot, sl2

usamoi:


VoVAllen:
Any performance number?

PR #615, opened 2024-11-19
release: v0.4.0

cutecutecat:
The update from `0.3.0` to `0.4.0` might fail!

Due to a conflict definition of `_vectors_svecf32_in` at `0.3.0`:
```
_vectors_svecf32_in(input, oid, _typmod)
```
and at 0.2.1--0.3.0:
```
_vectors_svecf32_in(input, oid, typmod)
```

As `0.4.0` picks latter signature, and PostgreSQL doesn't support [parameter name changing](https://stackoverflow.com/questions/72891280/is-it-possible-to-change-the-name-of-a-parameter-in-a-postgresql-function).

Therefore the update route of `0.2.1 -> 0.3.0 -> 0.4.0` can succeed, but `0.3.0 -> 0.4.0` would fail.

Ref: L744 of file `sql/upgrade/vectors--0.3.0--0.4.0.sql`

## Conclusion
Check and Keep these two signatures as their own.


usamoi:
> Due to a conflict definition of `_vectors_svecf32_in` at `0.3.0`:

Why? I cannot see. https://github.com/tensorchord/pgvecto.rs/blob/cc4776fec025672f934e3be6dfb2e27f930bef6d/sql/install/vectors--0.3.0.sql#L859. The parameter name is `_typmod`. Why do you introduce it?

sql/install/vectors--0.3.0.sql
```diff
@@ -859,7 +859,7 @@ AS 'MODULE_PATHNAME', '_vectors_svecf32_norm_wrapper';
 CREATE  FUNCTION "_vectors_svecf32_in"(
 	"input" cstring, /* &core::ffi::c_str::CStr */
 	"_oid" oid, /* pgrx_pg_sys::submodules::oids::Oid */
-	"typmod" INT /* i32 */
```
usamoi, line 862: It's strictly forbidden to edit historic schema scripts.

usamoi (changes requested):

sql/install/vectors--0.4.0.sql
```diff
@@ -0,0 +1,1967 @@
...
+
+-- src/datatype/subscript_vecf32.rs:10
+-- vectors::datatype::subscript_vecf32::_vectors_vecf32_subscript
+CREATE FUNCTION _vectors_vecf32_subscript(internal) RETURNS internal
+IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vectors_vecf32_subscript_wrapper';
+
+-- src/datatype/operators_vecf32.rs:109
+-- vectors::datatype::operators_vecf32::_vectors_vecf32_sphere_l2_in
+CREATE  FUNCTION "_vectors_vecf32_sphere_l2_in"(
+	"lhs" vector, /* vectors::datatype::memory_vecf32::Vecf32Input */
```
usamoi, line 35: remove all tabs

usamoi (approved):

PR #621, opened 2024-12-12
chore: add (S)SSE & NEON & SVE impl

usamoi:


VoVAllen:
You can use arm runner on ubicloud if needed. Ref: https://www.ubicloud.com/docs/github-actions-integration/migrate-to-arm-runners

PR #639, opened 2025-02-24
chore: fix ghcr release

kemingy:


ich777:
@kemingy thanks for pushing now to ghcr.io too. Would it be possible to push this tag to ghcr too: `tensorchord/pgvecto-rs:pg14-v0.2.0` since this would help users with Immich.

I assume this change was made because of the upcoming rate limit change on DockerHub on 1. April 2025

ich777:
@kemingy thanks, I saw that it synced about 10 minutes ago... :)
Sorry for the noise and thank you!

